PR #134, opened 2023-03-13
add server garbage collection; more sourceCount updating

### LINKED ISSUE #111: File Garbage Collection
When an image/chart/ramp config is removed from a slide via the editor, the link to the asset is removed from the Storyline config, but the associated file remains in it's folder. Would some sort of file garbage collection be desirable in this instance? This would prevent having unnecessary files pile up within the config file structure when creating/editing a Storyline product.

One immediate problem is that assets aren't unique, so if the same image was used in more than one place in the Storyline, deleting one would also affect the rest. This issue could be solved by assigning duplicate files some kind of uid, then storing them as separate files.

More trouble than its worth?

RyanCoulsonCA:
Closes #111 

This PR adds garbage collection to the server. When the ZIP is sent to the server, it will look at the files currently in the file system and compare them with the new files being uploaded through the ZIP (this should contain all of the up-to-date, active files) and will delete the files that aren't in the newly uploaded ZIP.

**I've done some thorough testing, but please do your best to see if you can come up with any edge cases or any way to break this since it does deal with deleting files and we don't want people to lose any work.**

## Changes

Editor Changes:
- When changing the right panel type, sources will be updated accordingly before clearing the panel config.
- Fix source counting not working correctly for slideshows (the counter was using the blob URL, not the actual file URL)
- When deleting an entire slide, panel sources will be updated correctly

Server Changes:
- Items removed from the product will now be removed from the server directory as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/134)

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -3,6 +3,7 @@ var bodyParser = require('body-parser'); //connects bodyParsing middleware
 var formidable = require('formidable');
 var path = require('path'); //used for file path
 var fs = require('fs-extra'); //File System-needed for renaming file etc
+var recursiveRead = require('recursive-readdir');
 var cors = require('cors');
 const decompress = require('decompress');
 const archiver = require('archiver');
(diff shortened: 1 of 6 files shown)


spencerwahl (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/134)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

___
*[`server/index.js` line 65 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/134#-NQVMz5a3k1kdtB6QXCN:-NQVMz5a3k1kdtB6QXCO:b-i97a5a) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/41744444c7c18e679ec81862ff83a29012f14fad/server/index.js#L65)):*
> ```JavaScript
>         // old files in the folder.
>         decompress(secureFilename, fileName).then((files) => {
>             // Below is some logic to remove items from the server directory that are no longer used.
> ```

In theory we could just wipe the directory since the used files will be replaced anyways... a little bit more worrying if things go wrong during the upload though so probably better to stick with this.

yileifeng (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/134)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

mohsin-r (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/134)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

Issue #152, opened 2023-03-27
Preview unsaved changes in new tab


yileifeng:
Ideally, opening the preview mode (for unsaved changes) should be redirected to a new tab to not disturb the user working on the current slide editor. This would involve figuring out how to preserve `configFileStructure` and `config` objects as well as its methods since data cannot be passed to a new tab locally without some type of storage.

Issue #129, opened 2023-03-12
"Back" button during Preview takes you back to landing page


dan-bowerman:
Using the "Back" button when previewing a Storylines product should take you back to where you were before the Preview, but instead takes you back to the landing page (Create New vs. Edit Existing). The expected behaviour would be to close the preview and resume where the user left off before clicking "Preview".

yileifeng:
Fix along with #89

Issue #94, opened 2023-02-21
Back to Home link


dan-bowerman:
A "Back to Home" link that takes you to the landing page (Create/Edit Storylines Product) would be useful. 

I am open to suggestions on where this could go. Right now, you must use the browser back button or change the URL manually.

yileifeng:
Related to #89

RyanCoulsonCA:
I'm looking for some feedback for how we might want to add this back button. I've included some options below and I'd love to hear what everyone thinks. 

### Option 1
We add a second row below the edit metadata button to return to the landing page.

![](https://i.gyazo.com/a6f3b472698cb80ae83a8991ee9783d4.png)


### Option 2
We add another button to the top of the page. We have quite a bit going on up there already, so maybe we don't want it here.
![](https://i.gyazo.com/14ee611c57aab74e33bdf8d6393cbb23.png)

### Option 3
Add a little exit icon to the left of the product name at the top.
![](https://i.gyazo.com/d9108d771b14d09e95c353911a13c244.png)

Let me know what you think, or if you have any other suggestions!

yileifeng:
Option 3 > Option 1 > Option 2 for me. Would just prefer to avoid over cluttering the right header section since the back button is not directly related to the current product like the other functions are.

RyanCoulsonCA:
> Option 3 > Option 1 > Option 2 for me. Would just prefer to avoid over cluttering the right header section since the back button is not directly related to the current product like the other functions are.

👍 these are my preferences too

PR #153, opened 2023-03-27
Fix Vue routing to separate slide editor from metadata editor

### LINKED ISSUE #89: Update editor routes
From @yileifeng in #84:

> Not exactly related to this PR but there is now a major routing issue I found while testing that should be opened as an urgent new issue. After exiting the landing page, we reach the /editor-metadata path but if we click next into the main editor, the path stays as /editor-metadata. The naming here is one issue since it doesn't make sense, but there are also bigger problems such as refreshing the page will take the user back to the main editor metadata page due to the same route. I think it also makes sense to have a back button or something of the sort to take the user back to the previous page, otherwise they are locked in the editor page unless they change the URL manually.
> 
> I think the best method is to make the main editor page (with the slides) as its own separate route (rather than keeping multiple templates in a single file) to clean up most of these issues, and maybe decide to rename the editor routes. Maybe there is a better solution I didn't think about though.

yileifeng:
Closes #89, #129, #169 

**Changes**: 
- added new routes: `/:lang/editor-main/:uid` for the slide editor (requires UID as parameter), `/:lang/editor-metadata/:uid` for metadata page with UID as parameter, `/:lang/editor/:uid` that redirects to the metadata page
- fixed back button in preview mode redirecting back to the slide editor (#129) but still should open in new tab  (opened #152)
- refactored code to separate metadata component and its logic from the main slide editor
- add functionality to load existing Storylines product for different page views if UID param is passed in route 
- renamed some files

**Testing**: 
Test transitioning between page views, save functionality, and loading products by passing in UID as URL parameter. Additionally, check all recently merged PR changes functionality in case any changes were lost during rebase.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153)

### DIFF
--- a/src/components/editor/dynamic-editor.vue
+++ b/src/components/editor/dynamic-editor.vue
@@ -152,16 +152,16 @@ export default class DynamicEditorV extends Vue {
     newSlideName = '';
     newSlideType = 'text';
 
-    get idUsed() {
+    get idUsed(): any {
         return this.panel.children.some((ch: any) => ch.id === this.newSlideName);
     }
 
-    changePanel(target: string) {
+    changePanel(target: string): void {
         this.saveChanges();
         this.editingStatus = target;
     }
 
-    switchSlide(idx: number) {
+    switchSlide(idx: number): void {
         // Save slide changes if neccessary and switch to the newly selected slide.
         this.saveChanges();
         this.editingSlide = idx;
@@ -175,7 +175,7 @@ export default class DynamicEditorV extends Vue {
         }
     }
 
-    removeSlide(item: any) {
+    removeSlide(item: any): void {
         const panel = this.panel.children.find((panel: any, idx: number) => idx === item).panel;
 
         // Update source counts based on which panel is removed.
@@ -215,7 +215,7 @@ export default class DynamicEditorV extends Vue {
         }
     }
 
-    createNewSlide() {
+    createNewSlide(): void {
         if (!this.newSlideName) return;
 
         const newConfig = {
@@ -227,7 +227,7 @@ export default class DynamicEditorV extends Vue {
         this.panel.children.push(newConfig);
     }
 
-    saveChanges() {
+    saveChanges(): void {
         if (this.$refs.slide !== undefined && typeof (this.$refs.slide as any).saveChanges === 'function') {
             (this.$refs.slide as any).saveChanges();
         }
--- a/src/components/editor/editor.vue
+++ b/src/components/editor/editor.vue
@@ -1,276 +1,168 @@
 <template>
     <!-- If the configuration file is being fetched, display a spinner to indicate loading. -->
     <div class="editor-container">
-        <template v-if="loadStatus === 'waiting' || loadStatus === 'loading'">
-            <div class="px-20 py-5">
-                <div class="flex">
-                    <div class="flex text-2xl font-bold mb-5">
-                        {{ editExisting ? $t('editor.editProduct') : $t('editor.createProduct') }}
-                    </div>
-                    <button @click="swapLang">
-                        {{ lang === 'en' ? $t('editor.frenchConfig') : $t('editor.englishConfig') }}
-                    </button>
-                </div>
-
-                <div class="border py-5 w-5/6">
-                    <label>{{ $t('editor.uuid') }}:</label>
-                    <input
-                        type="text"
-                        @input="error = false"
-                        v-model="uuid"
-                        class="w-1/3"
-                        :class="error ? 'input-error' : ''"
-                    />
-                    <button
-                        @click="generateRemoteConfig"
-                        class="bg-black text-white hover:bg-gray-800"
-                        :class="error ? 'input-error' : ''"
-                        v-if="editExisting"
-                    >
-                        {{ $t('editor.load') }}
-                    </button>
-
-                    <!-- If config is loading, display a small spinner. -->
-                    <div class="inline-flex" v-if="loadStatus === 'loading'">
-                        <spinner
-                            size="20px"
-                            background="#00D2D3"
-                            color="#009cd1"
-                            stroke="2px"
-                            class="mx-2 my-auto"
-                        ></spinner>
-                    </div>
-                </div>
-
-                <br />
-
-                <div class="mb-4">
-                    <h3>Storylines product details</h3>
-                    <p>
-                        Fill in metadata details about your new Storyline. Use the “Preview” button to see what your
-                        slides will look like.
-                    </p>
-                </div>
-
-                <metadata-editor
-                    :metadata="metadata"
-                    @metadata-changed="updateMetadata"
-                    @logo-changed="onFileChange"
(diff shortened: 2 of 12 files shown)


mohsin-r (changes requested):
Pretty neat changes! Three points that came to mind:

* This might be a bit nitpicky, but I'm not too sure that `metadata-editor.vue` is the right name for that file anymore since its doing a whole lot more than simply "editing the metadata".
* The unsaved changes indicator seems to be appearing as soon as you go into the main editor. For example, I'll save the product, preview, then go back to product, and it'll trigger the indicator before changes are made.
* Clicking on "View Saved Product" takes you to `/editor-main/editor-preview/Test`.

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-:-NRcu6OtC3RXlCbCbmgw:b-d6xkwu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yileifeng)

mohsin-r (commented):
Found another bug while testing. When you create a new product, preview it (without saving) and then click back, you get the UUID does not exist.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-:-NRdFN457EuNyEXsOxdO:b40tfbt)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yileifeng)

yileifeng (commented):
Open to any renaming suggestions for `metadata-editor` but left for now. Unsaved changes should be fixed as it was detecting the slides being initialized, added a timeout as couldn't find a better solution. The saved product and all transitions between routes have been reworked and should be fixed now (waiting on build update). 

For the preview mode, it actually does not work as is right now because the changes are not being saved before jumping to the preview mode, which means we are just loading in the saved product when we go back to the editor view. We could choose to save changes but this wouldn't make much sense as this function is for previewing unsaved changes. This is why I think the best solution is just to implement the new tab for preview mode #152, and this will also fix the bug you mentioned above. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-:-NRhvuaS0Pnusi8WkYDD:b-kbkx2s)* status: 5 of 9 files reviewed, all discussions resolved (waiting on @mohsin-r)

mohsin-r (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yileifeng)

RyanCoulsonCA (approved):
> Reviewed 5 of 8 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yileifeng)

yileifeng (commented):
Rebased changes from lang switching. There seems to be an issue reloading content after switching languages in the main slide editor so will open a new issue for that. Would be helpful if everyone could take a quick look and test to see if everything still functions as intended after the rebase. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-:-NSCN8jsELOYDElZyawl:bawsmkw)* status: 4 of 12 files reviewed, all discussions resolved (waiting on @mohsin-r and @RyanCoulsonCA)

mohsin-r (approved):
> Reviewed 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yileifeng)

spencerwahl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

___
*[`src/components/editor/metadata-editor.vue` line 534 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-NSCbRUVB0zgwlvdAH6Q:-NSCbRUVB0zgwlvdAH6R:b-dzxbbl) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/2a47efd3f343207243061ac9604967602d46f62d/src/components/editor/metadata-editor.vue#L534)):*
> ```Vue
>         this.lang = to.params.lang;
>         this.uuid = to.params.uid;
>         this.$i18n.locale = this.lang;
> ```

I think this function is the reason the language switching isn't working, `this.lang` is being used to denote which config is being edited but it is being used here as which lang the page is in.

roryhofland (approved):
Would we want the back button in the main editor to go back to the `/editor-metadata/uuid`endpoint instead of the landing page? Not a big deal either way, and if it causes extra issues with saving and whatnot then it's probably not worth the trouble.

Also very minor but when loading a config directly through a URL the 'loaded config', or 'uuid does not exist' message triggers twice. Guessing it's firing twice based on the way it loads the config then the editor, but I can take a look at this in #154 after this is merged.

> Reviewed 5 of 8 files at r1, 4 of 4 files at r2, 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-:-NRifSk00bkL2yxgtwCi:b-5iznem)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

yileifeng (commented):
For the back button, it shouldn't cause any more issues and I agree it makes sense to link it back to the `metadata/uid` route. The original issue https://github.com/ramp4-pcar4/storylines-editor/issues/94 seems to state a link back to the landing page though so unsure if we should make the change.

The second trigger was caused by the `beforeRouteUpdate` calling a second fetch to config when detecting route changes (don't believe this is required anymore). Removed and should fix this bug. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-:-NSGYH6e0e1CAQ7mWKKs:bihxfp3)* status: 10 of 12 files reviewed, 1 unresolved discussion (waiting on @mohsin-r, @roryhofland, and @spencerwahl)

___
*[`src/components/editor/metadata-editor.vue` line 534 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/153#-NSCbRUVB0zgwlvdAH6Q:-NSGZbPV7K4OrBfSLPyY:bc7o127) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/2a47efd3f343207243061ac9604967602d46f62d/src/components/editor/metadata-editor.vue#L534)):*

I think this function is the reason the language switching isn't working, `this.lang` is being used to denote which config is being edited but it is being used here as which lang the page is in.
</blockquote></details>

This should be fixed now, along with passing the config language as a separate prop to not disrupt the route param language.

Issue #164, opened 2023-04-03
Allow uploads that are unconnected to specific slides or panels


dan-bowerman:
Pie-in-the-sky stuff:

A user may want to embed an image in a text panel. It would be nice if there was an option somewhere to allow image uploads that could be referenced through markdown linking. Also related to #159, as these files would need to avoid getting deleted if they're not explicitly linked.

dan-bowerman:
Let's leave garbage collection out of this for the current phase, and revisit with a new issue in Phase 3.

PR #233, opened 2023-06-07
Inline Image Editor

roryhofland:
Related: #164 

Adds an inline image uploader to the markdown editor (**Upload Image** under the image icon).

Note: inline images do not appear in the markdown preview or storylines preview due to the way files are referenced in these pages. Obviously this is less than ideal, as it is much more helpful if the images appear when editing. I found that adding the storyline config with the uploaded assets to the `/public/` folder fixed this issue, and for some reason the issue remained fixed (the inline images were previewable) even after deleting the config from `/public/` again. Someone who is more knowledgeable with the file structure could offer some insight into this.

This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/233)

### DIFF
--- a/src/components/editor/text-editor.vue
+++ b/src/components/editor/text-editor.vue
@@ -8,13 +8,15 @@
             height="400px"
             left-toolbar="undo redo clear | h bold italic strikethrough quote subsuper | ul ol table hr | addLink image code | save"
             :toolbar="toolbar"
+            :disabled-menus="[]"
+            @upload-image="handleUploadImage"
         ></v-md-editor>
     </div>
 </template>
 
 <script lang="ts">
 import { Component, Prop, Vue } from 'vue-property-decorator';
-import { TextPanel } from '@/definitions';
+import { ConfigFileStructure, TextPanel } from '@/definitions';
 
 interface MDEditor {
     insert(callback: (selected: string) => { text: string; selected: string }): void;
@@ -25,6 +27,8 @@ interface MDEditor {
 })
 export default class TextEditorV extends Vue {
     @Prop() panel!: TextPanel;
+    @Prop() configFileStructure!: ConfigFileStructure;
+    @Prop() lang!: string;
 
     toolbar = {
         subsuper: {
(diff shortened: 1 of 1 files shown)


spencerwahl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/233)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @roryhofland)

___
*[`src/components/editor/text-editor.vue` line 123 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/233#-NXQRyUj86kiIMJmIKgB:-NXQRyUkDyv9r25AvDBb:b-h1gmx9) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/002020745ccd37f591f7c500174c345086367a17/src/components/editor/text-editor.vue#L123)):*
> ```Vue
>         this.configFileStructure.assets[this.lang].file(file.name, file);
> 
>         // Here is just an example
> ```

Was this a comment copied from an example? because the below thing looks like the proper code

RyanCoulsonCA (commented):
> Note: inline images do not appear in the markdown preview or storylines preview due to the way files are referenced in these pages.

Since the preview feature is intended to work without needing to save the product, it accesses images directly from the ZIP file ([see here](https://github.com/ramp4-pcar4/storylines-editor/blob/main/src/components/panels/slideshow-panel.vue#L92)). 

The library we're using to render markdown in the text panels (Markdown-it) just knows to display an image based on the source provided in the markdown, so it's trying to display an image with the path provided ( for example, `/00000000-0000-0000-0000-000000000000/assets/en/example.jpg`), but that path doesn't actually exist on the server yet and there isn't any logic to fetch the file from the ZIP so it just displays a broken image. 

[Markdown-it supports custom rendering](https://markdown-it.github.io/markdown-it/#Renderer.prototype.rules), so we can add some custom logic to the image rendering function to check whether the image exists in the ZIP file.

The problem with the markdown editor is probably the same as the Markdown-it problem, but the fix might be trickier because it doesn't seem like there is any way to add custom rendering. I don't have any good ideas for this one at the moment, if anyone else has any ideas please post!

I'm not sure how much work this involves, maybe it's worth a discussion for whether we want to do this now or push it to the next phase.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/233#-:-NXQaCC2Evb4XDlB-wAN:b5hg2ys)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @roryhofland and @spencerwahl)

dan-bowerman:
The issue here may be that the images need to be committed to the zip and not dumped to the public folder. Needs investigation.

yileifeng:
Much outdated and can reference code in open issue again if becomes requirement

PR #269, opened 2024-03-21
implement dynamic slideshow editor

RyanCoulsonCA:
### Related Item(s)
https://github.com/ramp4-pcar4/story-ramp/issues/410

### Changes
- Adds the dynamic slideshow editor.
- The existing `chart` and `image` editors still support uploading multiple charts and images, but their config is converted to a `slideshow` behind the scenes for a better user experience. The chart and image editors used in the dynamic slideshow editor only support one image or chart to prevent slideshow-ception.
- Removed some unused properties frmo `definitions.ts` 

### Notes
- I'm open to any and all UI changes for the dynamic panel. If you have an idea feel free to let me know!

### Testing
Steps:
1. Create a new slide with the type `slideshow` and play around with it.
2. The preview button will not currently work because of the schema changes (we need to merge the issue in the `story-ramp` issue first). You can verify that the changes are working by looking at the config files directly in the meantime!

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269)

### DIFF
--- a/src/components/editor/chart-editor.vue
+++ b/src/components/editor/chart-editor.vue
@@ -12,6 +12,7 @@
                 class="chart-btn bg-gray-100 cursor-pointer hover:bg-gray-200"
                 id="modal-btn"
                 @click="clearEditor()"
+                :disabled="!allowMany && chartConfigs.length > 0"
             >
                 <div class="flex items-center">
                     <svg height="18px" width="18px" viewBox="0 0 23 21" xmlns="http://www.w3.org/2000/svg">
@@ -44,6 +45,8 @@
                         :key="`${element.name}-${index}`"
                         :chart="element"
                         :configFileStructure="configFileStructure"
+                        :sourceCounts="sourceCounts"
+                        :lang="lang"
                         @edit="editChart"
                         @delete="$vfm.open(`${element.name}-${index}`)"
                     ></ChartPreview>
@@ -62,11 +65,18 @@
 
 <script lang="ts">
 import { Options, Prop, Vue } from 'vue-property-decorator';
-import { ChartConfig, ChartPanel, ConfigFileStructure, Highchart, SourceCounts } from '@/definitions';
+import {
+    ChartConfig,
+    ChartPanel,
+    ConfigFileStructure,
+    Highchart,
+    PanelType,
+    SlideshowPanel,
+    SourceCounts
+} from '@/definitions';
 import ChartPreviewV from '@/components/editor/helpers/chart-preview.vue';
 import ConfirmationModalV from '@/components/editor/helpers/confirmation-modal.vue';
 import draggable from 'vuedraggable';
-import { chart } from 'highcharts';
 
 @Options({
     components: {
@@ -78,10 +88,11 @@ import { chart } from 'highcharts';
     }
 })
 export default class ChartEditorV extends Vue {
-    @Prop() panel!: ChartPanel;
+    @Prop() panel!: ChartPanel | SlideshowPanel;
     @Prop() configFileStructure!: ConfigFileStructure;
     @Prop() lang!: string;
     @Prop() sourceCounts!: SourceCounts;
+    @Prop({ default: true }) allowMany!: boolean;
 
     edited = false;
 
@@ -106,9 +117,17 @@ export default class ChartEditorV extends Vue {
             );
         });
 
+        // This allows us to access the chart(s) using one consistent variable instead of needing to check panel type.
+        const charts =
+            this.panel.type === PanelType.Slideshow
+                ? (this.panel.items as Array<ChartPanel>)
+                : this.panel.src
+                ? [this.panel]
+                : [];
+
         // load charts from existing storylines product
-        if (this.panel.charts !== undefined && this.panel.charts.length) {
-            this.chartConfigs = this.panel.charts.map((chart: ChartConfig) => {
+        if (charts !== undefined && charts.length) {
+            this.chartConfigs = charts.map((chart: ChartPanel) => {
                 let chartName = '';
                 // extract chart name
                 if (chart.options && chart.options.title) {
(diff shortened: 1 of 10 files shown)


RyanCoulsonCA:
Because of the schema changes, existing products being built with RESPECT that contain chart or image panels will break. Do we think it's a good idea to add backwards-compatibility to the editor to automatically convert existing charts/image panels to the new schema? 

**Pros**:
- Would create less of a headache for the people working on existing products.
- It shouldn't be that much work to implement.

**Cons**:
- It kind of bloats our codebase. We could always remove it in the future though (give the clients a heads up they have x number of weeks/months to make the fixes, for example).
- Even though it shouldn't be much work, it's more work towards this issue that we've already spent a lot of time on

RyanCoulsonCA:
Demo site: https://ramp4-pcar4.github.io/storylines-editor/fix-410/#/en/editor

dan-bowerman:
> Do we think it's a good idea to add backwards-compatibility to the editor to automatically convert existing charts/image panels to the new schema?

The migration from the "Beta Editor" and its existing products will be a one-time event that will be happening hopefully soon. I was actually just talking about this with Mohsin regarding having similar functionality for upgrading RAMP3 map configs. As this functionality would only be needed once, we opted to not pursue it. 

What would need to happen to the charts and image panels to make them work? Would perhaps a standalone script for "upgrading" existing Storylines would suffice - perhaps just point it at a folder full of Storylines products and run? I don't think it's a good idea to bloat the RESPECT codebase for a one-time event.

RyanCoulsonCA:
The config changes required are outlined in the PR in the `story-ramp` repo (issue: https://github.com/ramp4-pcar4/story-ramp/pull/411).

> To fix existing chart galleries, change the type from `chart` to `slideshow`, change the `charts` property to `items`, and add `type: "chart"` to each of the entries (it now treats each of the charts as their own chart panel).

> To fix existing image slideshows, simply replace the `images` property with `items`.

These changes are simple enough that a standalone script should get the job done. If we want to bring our existing prod sites up to a new version we would need to run this script on those as configs well, obviously with backups just in case!

mohsin-r (commented):
I attempted to play around with it. I was able to successfully create a slideshow and I like the UI and functionality of it. However, when I tried previewing the product the slideshow didn't display and there are some console errors.
Another interesting point to consider with the slideshow is whether we can simply remove image, chart, and map types. Since the slideshow encompasses all cases of these individual panel types.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269#-:-NtXmuVjBHTQPPh2S904:bqbof29)* status: 0 of 11 files reviewed, all discussions resolved

RyanCoulsonCA (commented):
The preview wasn't working because we hadn't updated the NPM package to include the new dynamic slideshow changes in Storylines. I've updated the package now so it should be working! The maps are using a RAMP3 config so won't work until we merge in the RAMP4 editor.

For your second point, if I'm understanding it correctly, my main concern would be that I feel we would be taking a lot away from the user experience aspect. The reason why I added extra logic to keep all-image/all-chart slideshows on their respective editors instead of displaying the slideshow editor is that the user would need to do a lot more button pressing to add new charts, and images would have to be uploaded one at a time instead of just being able to drop a bunch in. Open to hearing the opinions of others on this!

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269#-:-Nta_col3LkguktQNM9Z:b-m2qg4z)* status: 0 of 14 files reviewed, all discussions resolved

yileifeng (commented):
I agree with keeping the individual image/chart/map editors for the flexibility and it does not seem intuitive for users to jump to the slideshow editor for simple panels such as a one image panel.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269#-:-NtbFKaA4boYSgV004vS:b8ij9lo)* status: 0 of 14 files reviewed, all discussions resolved

yileifeng (commented):
Problematic behavior appears when there are multiple charts in the dynamic slideshow:

The default file name of each chart is the same for each new chart editor that gets added, and this results in all charts in the slideshow pointing to the same file/instance (as opposed to adding multiple charts in the standalone chart editor which just has a integer incremented). Will need to figure out better default naming structure for unique file names.

In addition, renaming the chart file will cause more issues. This can be turned into a new issue, but we haven't really noticed the chart file naming to be much of an issue until now where it is magnified. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269#-:-NtbNSYJ8zQT5_Kcz4kO:b-oabxyv)* status: 0 of 14 files reviewed, all discussions resolved

RyanCoulsonCA (commented):
Made some changes:

- Enhanced our check to see if we're creating a chart with a duplicate name. It will now check to see if the new chart name already exists using the pre-existing `sourceCounts` object. If the file name is already in use, the user is alerted and the modal re-appears for the user to fix the issue.
- Instead of hiding the "add new chart" button, it is now disabled. There was an issue with having the v-if check on the same component that we had the modal connected to.
-  Fixed a bug where if you delete a slide while the editor is still open, nothing happens.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269#-:-Ntq_CWI0-rmRqY9FfY-:b-v1rd9)* status: 0 of 15 files reviewed, all discussions resolved

yileifeng (approved):
Chart issues seem to be resolved :+1:

> Reviewed 4 of 11 files at r1, 5 of 5 files at r2, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269#-:-Ntqh_4Z6t-rXHXD2wLx:bdg3qar)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

KashishMistry (approved):
> Reviewed 4 of 11 files at r1, 5 of 5 files at r2, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

szczz (approved):
> Reviewed 4 of 11 files at r1, 5 of 5 files at r2, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/269)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

Issue #331, opened 2024-06-20
Persistent warning everytime content type changes


dan-bowerman:
Found by Celina in 2024-06-18 Testing Party

## Description

This is more of an annoyance - if I create a new slide and try to change the content type, I get a warning before doing anything

## Environment

- I used Edge + Nonylphenol Storylines

## Steps to reproduce

1. Click to Nonylphenol storyline
2. Edit and add a new slide
3. Change content type

## Expected behavior

Because I've literally made no changes to content, I expect there to be no warning message

## Actual behavior
I get a warning message.
Again, this is just an annoying behavior.

PR #383, opened 2024-08-30
Hide panel change warnings when no changes made

gordlin:
### Related Item(s)
Issue #331 

### Changes
- [FEATURE] Hide "Panel changes will be deleted" warnings when there have been no changes made to the relevant panel. This affects two app interactions:
    - Changing the panel type via the `Content Type` dropdown
    - Changing the `Make the right panel the full slide` checkbox

### Testing
Steps:
1. Open any storylines product (I used `00000000-0000-0000-0000-000000000000`).
2. Click `New Slide` to add a new slide. Go to that new slide. 
3. Use the `Content Type` dropdown to change the panel type a few times. As long as you don't add any content, the panel type should change without a pop-up coming up.
4. Change to the slide to the "Text" type. Add something to the panel (e.g. change the panel title). Now try changing the panel type again; the popup should show up now.
5. Change to the right panel and try steps 3-4 again; results should be the same.
6. Click `New Slide` to add a new slide. Add some content to the right panel, then switch back to the left panel and try changing the panel type to "Dynamic". A warning message **should** pop up, as changing to dynamic overwrites both panels, not just the one you're on.
7. Click `New Slide` to add a new slide. Click the `Make the right panel the full slide` checkbox. As long as no changes are made to the **left** panel when there are both panels, there should be no pop-up warning. 
8. Try adding a change to the **left** panel and then clicking the checkbox; the pop-up should show up. (**Note**: The pop-up should now never show up when the left panel is hidden, unless there are changes and you are changing panel types).

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383)

### DIFF
--- a/src/components/chart-editor.vue
+++ b/src/components/chart-editor.vue
@@ -296,7 +296,7 @@ export default class ChartEditorV extends Vue {
 
     onChartsEdited(): void {
         this.edited = true;
-        this.$emit('slide-edit');
+        this.$emit('slide-edit', this.chartConfigs.length !== 0);
     }
 }
 </script>
--- a/src/components/dynamic-editor.vue
+++ b/src/components/dynamic-editor.vue
@@ -95,6 +95,7 @@
 import { Options, Prop, Vue } from 'vue-property-decorator';
 import {
     BasePanel,
+    BaseStartingConfig,
     ChartPanel,
     ConfigFileStructure,
     DefaultConfigs,
@@ -146,45 +147,7 @@ export default class DynamicEditorV extends Vue {
         video: 'video-editor'
     };
 
-    startingConfig: DefaultConfigs = {
-        text: {
-            type: PanelType.Text,
-            title: '',
-            content: ''
-        },
-        dynamic: {
-            type: PanelType.Dynamic,
-            title: '',
-            titleTag: '',
-            content: '',
-            children: []
-        },
-        slideshow: {
-            type: PanelType.Slideshow,
-            items: [],
-            userCreated: true
-        },
-        image: {
-            type: PanelType.Image,
-            src: ''
-        },
-        chart: {
-            type: PanelType.Chart,
-            src: ''
-        },
-        map: {
-            type: PanelType.Map,
-            config: '',
-            title: '',
-            scrollguard: false
-        },
-        video: {
-            type: PanelType.Video,
-            title: '',
-            videoType: '',
-            src: ''
-        }
-    };
+    startingConfig: DefaultConfigs = JSON.parse(JSON.stringify(BaseStartingConfig));
 
     editingStatus = 'text';
     editingSlide = -1;
--- a/src/components/image-editor.vue
+++ b/src/components/image-editor.vue
@@ -293,7 +293,7 @@ export default class ImageEditorV extends Vue {
 
     onImagesEdited(): void {
         this.edited = true;
-        this.$emit('slide-edit');
+        this.$emit('slide-edit', this.imagePreviews.length !== 0);
     }
 }
 </script>
--- a/src/components/slide-editor.vue
+++ b/src/components/slide-editor.vue
@@ -34,7 +34,13 @@
                             class="editor-input rounded-none cursor-pointer w-4 h-4"
                             v-model="rightOnly"
                             :disabled="rightOnly && determineEditorType(currentSlide.panel[panelIndex]) === 'dynamic'"
-                            @change.stop="$vfm.open(`right-only-${slideIndex}`)"
+                            @change.stop="
+                                if (currentSlide.panel.length > 1 && panelModified(currentSlide.panel[0])) {
+                                    $vfm.open(`right-only-${slideIndex}`);
+                                } else {
+                                    toggleRightOnly();
+                                }
+                            "
                         />
                         <span class="mx-2 font-bold">{{ $t('editor.slides.centerSlide') }}</span>
                         <input
@@ -236,8 +242,18 @@
                         <select
                             ref="typeSelector"
                             @input="
-                                $vfm.open(`change-slide-${slideIndex}`);
                                 newType = ($event.target as HTMLInputElement).value;
+                                if (
+                                    panelModified(currentSlide.panel[panelIndex]) ||
+                                    // Changing to dynamic deletes content of both panels. Therefore, warn if ANY panel has changes.
+                                    (newType === 'dynamic' && currentSlide.panel.some((panel) => panelModified(panel)))
+                                ) {
+                                    $vfm.open(`change-slide-${slideIndex}`);
+                                } else {
+                                    changePanelType(determineEditorType(currentSlide.panel[panelIndex]), newType);
+                                    toggleCenterPanel();
+                                    toggleCenterSlide();
+                                }
                             "
                             :value="determineEditorType(currentSlide.panel[panelIndex])"
                         >
@@ -269,7 +285,15 @@
                     :sourceCounts="sourceCounts"
                     :centerSlide="centerSlide"
                     :dynamicSelected="dynamicSelected"
-                    @slide-edit="$emit('slide-edit')"
+                    @slide-edit="(changedFromDefault: boolean = true) => {
+                        $emit('slide-edit');
+
+                        // changedFromDefault should hold a boolean indicating whether the panel is actually modified 
+                        // (different from initial state). Only needed for some multimedia editors; text editors
+                        // write directly to currentSlide constantly, which is handled by panelModified().
+                        currentSlide.panel[panelIndex].modified = changedFromDefault || undefined;
+                    }
+                    "
                     v-else
                 ></component>
             </div>
@@ -308,6 +332,7 @@
 import { Options, Prop, Vue, Watch } from 'vue-property-decorator';
 import {
     BasePanel,
+    BaseStartingConfig,
     ChartPanel,
     ConfigFileStructure,
     DefaultConfigs,
@@ -334,6 +359,7 @@ import SlideshowEditorV from './slideshow-editor.vue';
 import LoadingPageV from './helpers/loading-page.vue';
 import DynamicEditorV from './dynamic-editor.vue';
 import ConfirmationModalV from './helpers/confirmation-modal.vue';
+import { toRaw } from 'vue';
 
 @Options({
     components: {
@@ -383,13 +409,21 @@ export default class SlideEditorV extends Vue {
         this.currentSlide ? (this.rightOnly = this.currentSlide.panel.length === 1) : false;
     }
 
-    changePanelType(prevType: string, newType: string): void {
-        const startingConfig: DefaultConfigs = {
-            text: {
-                type: PanelType.Text,
-                title: '',
-                content: ''
-            },
+    /**
+     * Determines whether a given panel has been modified from the default configuration of its type.
+     * Note that some editors (e.g. text) write directly to currentSlide after each change,
+     * while other editors (e.g. image) do not. The first type is handled completely in
+     * panelModified; the second type requires you to set `panel.modified` for the given panel beforehand,
+     * indicating whether changes have been made from the specific editor sub-component (see
+     * `<component>`'s `@slide-edit` event handler).
+     * @param {BasePanel} panel The panel to analyze.
+     * @returns {boolean} Whether panel has been modified.
+     */
+    panelModified(panel: BasePanel): boolean {
+        const prevType = this.currentSlide.panel[this.panelIndex].type;
+
+        let startingConfig = {
+            ...JSON.parse(JSON.stringify(BaseStartingConfig)),
             dynamic: {
                 type: PanelType.Dynamic,
                 title:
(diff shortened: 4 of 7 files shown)


yileifeng (changes requested):
Right idea and the approach will work as long as `currentSlide` is updated, however this is not always the case for multimedia editors especially. For example, if you create a new slide with an image panel and upload an image, changing to a different editor type will no longer give the user the pop-up warning because the `currentSlide` was never updated (unless you navigate to a different panel or slide). This also applies to map, chart, slideshow, video editors. 

A few potential solutions for this is to always dynamically keep the `currentSlide` up to date whenever user makes content changes (like the text editor does). Alternatively, can attempt to update `currentSlide` or extract slide config from child editor component when user attempts to switch editor types and use that for `panelModified` comparison. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-:-O5tBzBj2jJJz1JfbVbC:b-ljowdq)* status: all files reviewed, 1 unresolved discussion (waiting on @gordlin)

___
*[`src/components/slide-editor.vue` line 441 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O5tD4j_53ZNvJMncOXL:-O5tD4j_53ZNvJMncOXM:b-rinfuf) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5cb35f9ace0df02dfa691bc865dc198365ed7a61/src/components/slide-editor.vue#L441)):*
> ```Vue
>                 config: `${this.configFileStructure.uuid}/ramp-config/${
>                     this.configFileStructure.uuid
>                 }-map-${this.getNumberOfMaps()}.json`,
> ```

Bit confused about the starting map config property, shouldn't this be an empty string? I see for other `startingConfig` instances across the app this is the case. Similarly, can we just set title and content properties to empty strings for dynamic panel configs above (unless there is a specific use case this breaks)?

On that note, I think it is best to move the `startingConfig` object to the `definitions.ts` file and import it in the components where it is needed as there are multiple instances using the same/similar `startingConfig` object (if there are minor differences can just create a separate local component variable and modify the property needed). If that can be done in this PR would be great code cleanup.

gordlin (commented):
I've fixed these issues, and also all remaining edge cases (that I know of).

There should now also be no pop-up for these cases:
- Selecting an image in the image editor and then deleting it
-  Adding slides in the slide editor and deleting them all
    - Strange case in slideshow editor, where selecting a map slide ensured there would always be a pop-up, even after the above was fixed
- Adding chart in the chart editor and then deleting it
- Modifying any field in the video editor and then erasing all changes

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-:-O66_Idb6lUUPbMjovYA:bd39qvv)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @yileifeng)

___
*[`src/components/slide-editor.vue` line 441 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O5tD4j_53ZNvJMncOXL:-O66a5he3WlUOqfaXRHy:b2etu9k) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5cb35f9ace0df02dfa691bc865dc198365ed7a61/src/components/slide-editor.vue#L441)):*

Bit confused about the starting map config property, shouldn't this be an empty string? I see for other `startingConfig` instances across the app this is the case. Similarly, can we just set title and content properties to empty strings for dynamic panel configs above (unless there is a specific use case this breaks)?

On that note, I think it is best to move the `startingConfig` object to the `definitions.ts` file and import it in the components where it is needed as there are multiple instances using the same/similar `startingConfig` object (if there are minor differences can just create a separate local component variable and modify the property needed). If that can be done in this PR would be great code cleanup.
</blockquote></details>

Refactored.

gordlin (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @yileifeng)

___
*[`src/components/slide-editor.vue` line 441 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O5tD4j_53ZNvJMncOXL:-O68Gu0g8X8vf1OMf87G:bj7p2su) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5cb35f9ace0df02dfa691bc865dc198365ed7a61/src/components/slide-editor.vue#L441)):*
<details><summary><i>Previously, gordlin (Gordon Lin) wrote…</i></summary><blockquote>

Refactored.
</blockquote></details>

Additional note, as I forgot to address the first half: I believe the extra map `config` stuff is necessary here. When we switch to the map panel type, the `createNewConfig()` function inside the map editor is called, which changes the `config` value from a blank string to this. Comparing the panel to one with a blank `config` value would thus make this checker think there've been changes when there haven't. 

There may be a similar issue with the dynamic configs.

yileifeng (changes requested):
Behavior is working properly for all cases I tested :+1:

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-:-O67Y0XE5CFJCXO4hrmF:b509hvc)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @gordlin)

___
*[`src/definitions.ts` line 287 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67_ftI-lBWLtzqBYGO:-O67_ftI-lBWLtzqBYGP:bb8sa2v) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/definitions.ts#L287)):*
> ```XML
> }
> 
> export const baseStartingConfig: DefaultConfigs = {
> ```

capitalize the naming for consistency
___
*[`src/components/slide-editor.vue` line 41 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67a70k--yMy7-UQbUi:-O67a70k--yMy7-UQbUj:bmptowg) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L41)):*
> ```Vue
>                                     $vfm.open(`right-only-${slideIndex}`);
>                                 } else {
>                                     // currentSlide.panel[panelIndex].modified = false;
> ```

Remove commented out line if unneeded
___
*[`src/components/slide-editor.vue` line 289 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67YRd-4xYP0n7KymCD:-O67YRd-4xYP0n7KymCE:bqo13lt) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L289)):*
> ```Vue
>                     :centerSlide="centerSlide"
>                     :dynamicSelected="dynamicSelected"
>                     @slide-edit="(args: any) => {
> ```

Code block is a bit long so should go inside a new component method. For the parameter, can just use `madeChanges: boolean` (or some other descriptive name) instead of `args: any`. Can we set the param default to `True` to remove the undefined check as this is the default/else behavior? More on that below.
___
*[`src/components/slide-editor.vue` line 295 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O6MIHSjE9vKOBDYF21T:-O6MIHSjE9vKOBDYF21U:b-g8vahc) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L295)):*
> ```Vue
>                             // (different from initial state). Only needed for some multimedia editors; text editors
>                             // write directly to currentSlide constantly, which is handled by panelModified().
>                             switch(currentSlide.panel[panelIndex].type) {
> ```

I think I understand the idea, but it would be much much cleaner code to not add a switch/case statement for a subset of panels. I assume the default case is what gets processed for text and dynamic panels? In that case, maybe defaulting the boolean parameter to `True` would work. An alternative approach is to pass the panel config as a parameter when emitting and run that against the `panelModified` helper. Ideally, we just want a general method to normalize behavior of when slides are edited for different panel types and remove the switch statement which should be doable. 
___
*[`src/components/slide-editor.vue` line 300 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O6MJU9YCk_pblWk8-OY:-O6MJU9YCk_pblWk8-OZ:b-r1ewia) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L300)):*
> ```Vue
>                                 case 'video':
>                                 case 'slideshow':
>                                     currentSlide.panel[panelIndex].modified = (args as boolean) ? true : undefined;
> ```

Could be misunderstanding, but should this be something like `if (args) { currentSlide.panel[panelIndex].modified = args }` (with a better variable name for `args`)?
___
*[`src/components/slide-editor.vue` line 326 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67sRny7g1C6Ylob_yz:-O67sRny7g1C6Ylob_z-:byp18fd) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L326)):*
> ```Vue
>             "
>             @ok="
>                 // currentSlide.panel[panelIndex].modified = false;
> ```

Remove

gordlin (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @yileifeng)

___
*[`src/definitions.ts` line 287 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67_ftI-lBWLtzqBYGO:-O6NOfDl2RHh3x-Jbny9:b-ugpthr) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/definitions.ts#L287)):*

capitalize the naming for consistency
</blockquote></details>

Donethanks!
___
*[`src/components/slide-editor.vue` line 41 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67a70k--yMy7-UQbUi:-O6NOhjx41bm7bKtpCC-:b-ugpthr) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L41)):*

Remove commented out line if unneeded
</blockquote></details>

Donethanks!
___
*[`src/components/slide-editor.vue` line 289 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67YRd-4xYP0n7KymCD:-O6NOm7xCBvtfZVX8FeM:b-yrhdyp) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L289)):*

Code block is a bit long so should go inside a new component method. For the parameter, can just use `madeChanges: boolean` (or some other descriptive name) instead of `args: any`. Can we set the param default to `True` to remove the undefined check as this is the default/else behavior? More on that below.
</blockquote></details>

Block's been shortened to one line, so it shouldn't need a new method anymore.
___
*[`src/components/slide-editor.vue` line 295 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O6MIHSjE9vKOBDYF21T:-O6NOtXBD8rDwBwR0lMB:b-2s9esa) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L295)):*

I think I understand the idea, but it would be much much cleaner code to not add a switch/case statement for a subset of panels. I assume the default case is what gets processed for text and dynamic panels? In that case, maybe defaulting the boolean parameter to `True` would work. An alternative approach is to pass the panel config as a parameter when emitting and run that against the `panelModified` helper. Ideally, we just want a general method to normalize behavior of when slides are edited for different panel types and remove the switch statement which should be doable.
</blockquote></details>

Donethanks, refactored into a single line.
___
*[`src/components/slide-editor.vue` line 300 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O6MJU9YCk_pblWk8-OY:-O6NPEIb3mxLDt765_WE:b-9ukzpw) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L300)):*

Could be misunderstanding, but should this be something like `if (args) { currentSlide.panel[panelIndex].modified = args }` (with a better variable name for `args`)?
</blockquote></details>

Setting `args`/`changedFromDefault` to `false` messes with `panelModified()`'s check whether the panel's different from its default configuration object (`{}` is different from `{property:  false}`). Setting it to `undefined` prevents this issue.

I've refactored this bit of code to be cleaner (use `||` instead of ternary).
___
*[`src/components/slide-editor.vue` line 326 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O67sRny7g1C6Ylob_yz:-O6NPKl89XPWFFaubOM6:b-ugpthr) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/f83d2ee3c6320139a9c37059889406fc155585b2/src/components/slide-editor.vue#L326)):*

Remove
</blockquote></details>

Donethanks!

yileifeng (approved):
> Reviewed 3 of 7 files at r2, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @gordlin)

IshavSohal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383)* status: all files reviewed, 1 unresolved discussion (waiting on @gordlin)

*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O6bkIW9EE864grmfouX:-O6bkIW9EE864grmfouY:bcejojk) (no related file):*
Noticed that when you create a new slide, change the content type to slideshow, add text to the title and then remove it, changing the content type again (to something other than dynamic) brings up the warning. 

All other cases appear to be functioning correctly though. 

___

gordlin (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @IshavSohal and @yileifeng)

*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383#-O6bkIW9EE864grmfouX:-O6wphYdCuwXDEAt1Sd1:b-ugpthr) (no related file):*

Noticed that when you create a new slide, change the content type to slideshow, add text to the title and then remove it, changing the content type again (to something other than dynamic) brings up the warning. 

All other cases appear to be functioning correctly though.
</blockquote></details>

Donethanks!

___

IshavSohal (approved):
> Reviewed 3 of 7 files at r2, 2 of 4 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/383)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @gordlin)

Issue #385, opened 2024-09-04
Investigate and implement changes for a unified assets folder


szczz:
## Description
Currently there are separate EN and FR folders within the assets folder. This means there are often duplicate assets that result in bloated Storyline sizes. This can have a significant performance impact especially now that we save the last 5 versions of a Storylines product. 

We need to investigate and implement a solution that allows both EN and FR storylines to access a single file in the asset folder but also enable each version to have their own unique assets if required.

## Related issues

#213

Issue #343, opened 2024-06-20
Ghost images in zip container


dan-bowerman:
Found by Matt in 2024-06-18 Testing Party

Adding an image and then changing to another slide before saving results in the image being lost. The name of the image will persist in the zip though. Likely part of the issue clients have been reporting on UAT.

RyanCoulsonCA:
> Adding an image and then changing to another slide before saving results in the image being lost.

This should be fixed after https://github.com/ramp4-pcar4/storylines-editor/pull/356 is merged, but we should leave this issue open until we confirm this is the case.

If a user uploads an image and then saves the product, the image will persist on the server even if it's been removed from the product because we don't have any garbage collection in place on the server side to remove unused files. We've solved this in the past but ran into some issues so we reverted it (see https://github.com/ramp4-pcar4/storylines-editor/issues/159).

RyanCoulsonCA:
The removed garbage collection code can be found here: https://github.com/ramp4-pcar4/storylines-editor/pull/165/files#diff-bcb729747c92e72c12e4590f736a334041a572c429756da8af60b6ab89c6be79

Issue #213, opened 2023-05-11
Copying slides between languages should copy to appropriate language folders as well


dan-bowerman:
Related #70.

Currently, if you use the "Copy slides from a different language" function, it correctly copies those slide contents to the French JSON structure.

However, images, maps, and charts do not appear to get copied. The JSON structure will still reference the `en` folders for `assets`, `charts`, `ramp-config`. These should get copied to the `fr` folder, and references in the French JSON structure (or opposite language), should be updated to reflect `fr` file paths. 

![Image](https://github.com/ramp4-pcar4/storylines-editor/assets/5294596/ed58e4af-8cfe-4188-a11f-56c63be993e2)

dan-bowerman:
What if we copied the slide structure, but not the files - and then in another issue, created the ability to know if assets and references are missing? Then the clients can just replace the missing items in the French side of the editor.

PR #414, opened 2024-10-23
Create shared asset system

IshavSohal:
### Related Item(s)
#385 
#213 

### Changes
- Added a shared assets folder. 
    - When an asset is uploaded within the config of one lang that already exists in the other lang's config, it gets removed from the other lang's assets folder, and added to the shared assets folder
    - When an asset is removed from a lang's assets folder, the asset in the current lang's assets folder has its source count set to 0, and it is added onto to the source count of the asset in the shared assets folder. This is because all instances of that asset within the current lang's config will be set to refer to the shared assets folder (once `save changes` is clicked, that is)
- Modified `deleteVideo()` to decrement the source count of video asset (if its local), and to remove it from `configFileStructure`
- Added a `shared-asset` event, which takes as input the name of an asset
    - This event is emitted when an asset uploaded to the current lang's config already exists in the opposite lang's config, and thus needs to be moved to the shared assets folder
    - When this event is emitted, it modifies the src of each instance of the uploaded asset (within the opposite lang's config) to refer to the shared assets folder
- Created methods for comparing contents of files, by comparing their hashes
- Created a `panelHelper`  method that executes a provided callback function on each ImagePanel and VideoPanel within a specified panel
- Incremented the source count of all assets within a slide when it is copied
- Moved all assets within a slide to the shared folder when its config is copied to the opposite lang

### Notes
- When deleting an asset and its source count hits 0, it gets removed from `configFileStructure`. However the asset doesn't get removed from the assets folder of the product. Due to this, when the app is reloaded and `configFileStructure` is initialized, the deleted asset returns. If this is a bug, I can open a new issue for this. 
- Once this PR is merged, I will try to get the garbage collection (from #343) working again, so that assets are appropriately removed from the product folder itself
- console logs are just for testing, will be removed before merging

### Testing

Perform the below steps using video panels as well. It should also work in the same way

#### Creating product
1. Create a new storylines product and fill in necessary metadata. Be sure to include a logo
2. Click Next

#### Adding an asset to configs of both lang's
3. Within the `en` config, create a new slide, and then an image panel within it
4. Create an image panel within it
5. Upload an image
6. Observe, in the console, that `configFileStrucuture` stores the asset within the `assets/en` folder
7. Click `save changes`
8. Switch to the `fr` config
9. Create new slide, and then an image panel within it
10. Upload the exact same image that you uploaded in the `en` config
11. Observe, in the console, that `configFileStrucuture` stores the asset within the `assets/shared` folder, and has removed it from the `assets/en` folder
12. Click `save changes`
13. Check the config file for each language, and observe that the src of the asset now refers to the shared assets folder

#### Deleting an asset with a source count greater than 1
14. Within the `en` config, delete the image that you uploaded
15. Observe, in the console, that the image still exists in `configFileStructure` within the `assets/shared` folder (since its source count has gone from 2 to 1)
16. Click `save changes` 

#### Deleting an asset with a source count of 1
17. Switch over to the `fr` config
18. Delete the image that you uploaded here
19. Observe, in the console, that the image now doesn't exists within the `assets/shared` folder (since its source count has gone from 1 to 0)
20. Click `save changes`

#### Changing the type of a panel that contains assets from shared assets folder
21. Upload a (new) image to the `en` and `fr` configs
22. Click `save changes`
23. Within one config, switch the panel type of the image panel
24. Observe, in the console, that the image still exists within the `assets/shared` folder (since its source count has gone from 2 to 1)
25. Observe, in the advanced editor, that the image uploaded now doesn't exist
26. Click `save changes`
27. Switch to the other config, and switch the panel type of the image panel
28.  Observe, in the console, that the image now doesn't exists within the `assets/shared` folder (since its source count has gone from 1 to 0)
29. Observe, in the advanced editor, that the image uploaded now doesn't exist
30. Click `save changes`

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)

### DIFF
--- a/src/components/dynamic-editor.vue
+++ b/src/components/dynamic-editor.vue
@@ -92,6 +92,9 @@
                     :centerSlide="centerSlide"
                     :dynamicSelected="dynamicSelected"
                     @slide-edit="$emit('slide-edit')"
+                    @shared-asset="(oppositeAssetPath: string, sharedAssetPath: string, oppositeLang: string) => {
+                        $emit('shared-asset', oppositeAssetPath, sharedAssetPath, oppositeLang);
+                    }"
                 ></component>
             </div>
         </div>
--- a/src/components/editor.vue
+++ b/src/components/editor.vue
@@ -242,6 +242,14 @@
                     @slide-change="selectSlide"
                     @slides-updated="updateSlides"
                     @open-metadata-modal="$vfm.open('metadata-edit-modal')"
+                    @shared-asset="(oppositeAssetPath: string, sharedAssetPath: string, oppositeLang: string) => {
+                        $emit('shared-asset', oppositeAssetPath, sharedAssetPath, oppositeLang);
+                    }"
+                    @process-panel="
+                        (panel, callback, ...args) => {
+                            $emit('process-panel', panel, callback, ...args);
+                        }
+                    "
                     :configFileStructure="configFileStructure"
                     :lang="configLang"
                     :sourceCounts="sourceCounts"
@@ -259,6 +267,14 @@
                     @slides-updated="updateSlides"
                     @open-metadata-modal="$vfm.open('metadata-edit-modal')"
                     @close-sidebar="closeSidebar"
+                    @shared-asset="(oppositeAssetPath: string, sharedAssetPath: string, oppositeLang: string) => {
+                        $emit('shared-asset', oppositeAssetPath, sharedAssetPath, oppositeLang);
+                    }"
+                    @process-panel="
+                        (panel, callback, ...args) => {
+                            $emit('process-panel', panel, callback, ...args);
+                        }
+                    "
                     :configFileStructure="configFileStructure"
                     :lang="configLang"
                     :sourceCounts="sourceCounts"
@@ -278,6 +294,9 @@
                     :slideIndex="slideIndex"
                     :isLast="slideIndex === slides.length - 1"
                     :uid="uuid"
+                    @shared-asset="(oppositeAssetPath: string, sharedAssetPath: string, oppositeLang: string) => {
+                        $emit('shared-asset', oppositeAssetPath, sharedAssetPath, oppositeLang);
+                    }"
                     @slide-change="selectSlide"
                     @slide-edit="onSlidesEdited"
                     @custom-slide-updated="updateCustomSlide"
@@ -312,14 +331,17 @@
 <script lang="ts">
 import { Options, Prop, Vue, Watch } from 'vue-property-decorator';
 import {
+    BasePanel,
     ConfigFileStructure,
     HelpSection,
+    ImagePanel,
     MetadataContent,
     MultiLanguageSlide,
     Slide,
     SourceCounts,
     StoryRampConfig,
-    TextPanel
+    TextPanel,
+    VideoPanel
 } from '@/definitions';
 import { VueSpinnerOval } from 'vue3-spinners';
 import axios from 'axios';
@@ -448,15 +470,19 @@ export default class EditorV extends Vue {
             panel: [{ type: 'loading-page' }, { type: 'loading-page' }]
         };
 
+        const newLang = lang || this.configLang || 'en';
+        if (this.configLang !== newLang) {
+            this.$emit('lang-change', newLang);
+        }
+
         setTimeout(() => {
             if (index === -1 || !this.loadSlides) {
                 this.currentSlide = '';
             } else {
-                const selectedLang = (lang ?? this.configLang) as keyof MultiLanguageSlide;
+                const selectedLang = newLang as keyof MultiLanguageSlide;
                 const selectedSlide = this.loadSlides[index][selectedLang];
                 this.currentSlide = selectedSlide ?? '';
             }
-
             this.slideIndex = index;
             (this.$refs.slide as SlideEditorV).panelIndex = 0;
             (this.$refs.slide as SlideEditorV).advancedEditorView = false;
--- a/src/components/helpers/image-preview.vue
+++ b/src/components/helpers/image-preview.vue
@@ -17,7 +17,7 @@
             <div class="flex-grow image-container">
                 <img
                     class="image-file object-cover"
-                    :title="imageFile.id"
+                    :title="imageFile.id.split('/').at(-1)"
                     :src="imageFile.src"
                     :alt="imageFile.altText"
                 />
--- a/src/components/image-editor.vue
+++ b/src/components/image-editor.vue
@@ -152,6 +152,7 @@ import { Options, Prop, Vue } from 'vue-property-decorator';
 import { ConfigFileStructure, ImageFile, ImagePanel, PanelType, SlideshowPanel, SourceCounts } from '@/definitions';
 import draggable from 'vuedraggable';
 import ImagePreviewV from './helpers/image-preview.vue';
+import JSZip from 'jszip';
 
 @Options({
     components: {
@@ -208,34 +209,24 @@ export default class ImageEditorV extends Vue {
                 // Check if the config file exists in the ZIP folder first.
                 const assetSrc = `${image.src.substring(image.src.indexOf('/') + 1)}`;
                 const filename = image.src.replace(/^.*[\\/]/, '');
-                const assetFile = this.configFileStructure.zip.file(assetSrc);
+                const compressedAssetFile = this.configFileStructure.zip.file(assetSrc);
                 const assetType = assetSrc.split('.').at(-1);
 
-                if (assetFile) {
-                    if (assetType != 'svg') {
-                        this.imagePreviewPromises.push(
-                            assetFile.async('blob').then((res: Blob) => {
-                                return {
-                                    ...image,
-                                    id: filename ? filename : image.src,
-                                    src: URL.createObjectURL(res)
-                                } as ImageFile;
-                            })
-                        );
-                    } else {
-                        this.imagePreviewPromises.push(
-                            assetFile.async('text').then((res) => {
-                                const imageFile = new File([res], filename, {
+                if (compressedAssetFile) {
+                    this.imagePreviewPromises.push(
+                        compressedAssetFile.async(assetType !== 'svg' ? 'blob' : 'text').then((assetFile) => {
+                            if (assetType === 'svg') {
+                                assetFile = new File([assetFile], filename, {
                                     type: 'image/svg+xml'
                                 });
-                                return {
-                                    ...image,
-                                    id: filename ? filename : image.src,
-                                    src: URL.createObjectURL(imageFile)
-                                } as ImageFile;
-                            })
-                        );
-                    }
+                            }
+                            return {
+                                ...image,
+                                id: image.src,
+                                src: URL.createObjectURL(assetFile)
+                            } as ImageFile;
+                        })
+                    );
                 }
             });
 
@@ -244,36 +235,210 @@ export default class ImageEditorV extends Vue {
                 this.imagePreviews = res;
                 this.imagePreviewsLoading = false;
             });
-
             this.slideshowCaption = this.panel.caption as string;
         }
+    }
+
+    // TODO: move method into a plugin. That way it isn't repeated in the metadata/video editors
+    // Converts a file into a promise that resolves to an ArrayBuffer containing the files data
+    readBinaryData(file: File): Promise<ArrayBuffer> {
+        return new Promise((resolve, reject) => {
+            const fileReader = new FileReader();
+            fileReader.onload = () => {
+                resolve(fileReader.result);
+            };
+            fileReader.onerror = () => {
+                reject(new Error('Could not load file reader'));
+            };
+            fileReader.readAsArrayBuffer(file);
+        });
+    }
+
+    // TODO: move method into a plugin. That way it isn't repeated in the metadata/video editors
+    // Converts a file into a promise that resolves to its has, as an array of 8-bit integers
+    obtainHashData(file: File): Promise<Uint8Array> {
+        return this.readBinaryData(file)
+            .then((res) => {
+                res = new Uint8Array(res);
+                return window.crypto.subtle.digest('SHA-256', res);
+            })
+            .then((res) => {
+                res = new Uint8Array(res);
+                return res;
+            });
+    }
+
+    // TODO: move method into a plugin. That way it isn't repeated in the metadata/video editors
+    /**
+     * Helper used to find all instances of the specified file in the specified asset folder
+     * @param file File that was uploaded
+     * @param folder The asset folder withinn which we should be searching
+     * @param checkNested Flag that indicates whether we should consider assets with nested subfolders
+     */
+    filesInAssetFolder(file: File, folder: string, checkNested = true): Array<Promise<string>> {
+        // Here, if a file in the specified folder has the same name and hash as the file uploaded, then we consider the
+        // two to be the same. Otherwise, we consider them to be different. We even consider if the asset is within a
+        // subfolder of the specified folder, so long as the name and hash of the file is the same. There may be more than one
+        // instance of the specified asset in the specified folder, albeit in seperate subfolders, hence why we collect
+        // an array of duplicate asset promises
+        const sharedAssetPromises = [];
+        this.configFileStructure.assets[folder].forEach((relativePath, compressedBinary) => {
+            const assetName = checkNested ? relativePath.split('/').at(-1) : relativePath;
+
+            // Only compare the hashes of two files when they have the exact same name
+            if (assetName === file.name) {
+                sharedAssetPromises.push(
+                    this.compareFiles(file, compressedBinary, assetName).then((fileSame) =>
+                        fileSame ? relativePath : 'N/A'
+                    )
+                );
+            }
+        });
+
+        return sharedAssetPromises;
+    }
+
+    // TODO: move method into a plugin. That way it isn't repeated in the metadata/video editors
+    /**
+     * Compares the hashes of two files
+     * @param file File that was uploaded
+     * @param compressedBinary Compressed binary file from configFileStructure
+     * @param compressedName The name of the compressed binary file
+     */
+    async compareFiles(file: File, compressedBinary: JSZip.JSZipObject, compressedName: string): Promise<boolean> {
+        const fileHash = await this.obtainHashData(file);
+        const compressedType = compressedName.split('.').at(-1);
+
+        return compressedBinary
+            .async(compressedType !== 'svg' ? 'blob' : 'text')
+            .then((assetFile) => {
+                if (compressedType === 'svg') {
+                    assetFile = new File([assetFile], compressedName, {
+                        type: 'image/svg+xml'
+                    });
+                }
+                return this.obtainHashData(assetFile);
+            })
+            .then((hash) => {
+                return hash.join() === fileHash.join();
+            });
+    }
+
+    // TODO: move method into a plugin. That way it isn't repeated in the metadata/video editors
+    // Helper for onFileChange and dropImages. Maps a File object to an ImageFile object
+    async addUploadedFile(file: File): Promise<ImageFile> {
+        const oppositeLang = this.lang === 'en' ? 'fr' : 'en';
+        const sharedAssetPaths = await Promise.all(this.filesInAssetFolder(file, 'shared', false));
+        let inSharedAsset = false;
+        let oppositeSourceCount = 0;
+        let newAssetName = file.name;
+        let uploadSource = `${this.configFileStructure.uuid}/assets/shared/${file.name}`;
+
+        // Should contain either 0 or 1 promise.
+        sharedAssetPaths.forEach((sharedAssetPath) => {
+            inSharedAsset = sharedAssetPath !== 'N/A';
+        });
+
+        // If the asset is already in the shared asset folder, we do not need to do anything. We can assume that it
+        // does not exist in the opposite lang's asset folder (i.e. we assume shared asset system is working correctly)
+        if (!inSharedAsset) {
+            const oppositeAssetPaths = await Promise.all(this.filesInAssetFolder(file, oppositeLang));
+            // If the current promise contains 'N/A', then the current path refers to an asset in the opposite asset
+            // folder that has the same name, but different contents, as the asset uploaded. In this case we do
+            // nothing, as this asset is not a valid duplicate.
+            for (const oppositeAssetPath of oppositeAssetPaths) {
+                if (oppositeAssetPath !== 'N/A') {
+                    const oppositeFileSource = `${this.configFileStructure.uuid}/assets/${oppositeLang}/${oppositeAssetPath}`;
+                    oppositeSourceCount += this.sourceCounts[oppositeFileSource] ?? 0;
+                    this.sourceCounts[oppositeFileSource] = 0;
+                    this.configFileStructure.assets[oppositeLang].remove(oppositeAssetPath);
+
+                    // Add asset to shared folder if asset is yet to be moved to the shared folder. If an asset with the
+                    // same name, but different content, is already in the shared folder, we must give the asset we are
+                    // uploading a unique name. Otherwise the existing asset will be overwritten
+                    if (!inSharedAsset) {
+                        let i = 2;
+                        while (this.configFileStructure.assets['shared'].file(newAssetName)) {
+                            // If the updated name is the same as a file that already exists in the shared asset folder,
+                            // we must compare that file with the uploaded file, since they wouldnt have been compared
+                            // on the first run due to having different names (Such an asset COULD also be in the shared
+                            // or opposite lang asset folder, since a new name is now being used. However we shouldn't
+                            // execute this entire method again just for an edge case like a name change, so we can
+                            // leave this as is for now)
+                            if (i > 2) {
+                                const filesEqual = await this.compareFiles(
+                                    file,
+                                    this.configFileStructure.assets['shared'].file(newAssetName),
+                                    newAssetName
+                                );
+                                if (filesEqual) break;
+                            }
+                            newAssetName = `${file.name.split('.').at(0)}(${i}).${file.name.split('.').at(-1)}`;
+                            i++;
+                        }
+                        uploadSource = `${this.configFileStructure.uuid}/assets/shared/${newAssetName}`;
+                        this.configFileStructure.assets['shared'].file(newAssetName, file);
+                        inSharedAsset = true;
+                    }
+                    this.$emit('shared-asset', oppositeFileSource, uploadSource, oppositeLang); // must be emitted for each duplicate asset
+                }
+            }
+        }
+
+        // If the asset uploaded is in the shared asset folder, then no need to upload to the current langs assets folder
+        if (!inSharedAsset) {
+            const currAssetPaths = await Promise.all(this.filesInAssetFolder(file, this.lang, false));
+            // Should contain either 0 or 1 promise.
+            // Need to use await here
+            for (const currAssetPath of currAssetPaths) {
+                // If asset w/ same name but different contents is in curr lang asset folder, set name in curr lang
+                // asset folder to a unique name, to avoid overwriting an existing file.
+                if (currAssetPath === 'N/A') {
+                    let i = 2;
+                    while (this.configFileStructure.assets[this.lang].file(newAssetName)) {
+                        // If the updated name is the same as a file that already exists in the current langs asset folder,
+                        // we must compare that file with the uploaded file, since they wouldnt have been compared
+                        // on the first run due to having different names
+                        if (i > 2) {
+                            const filesEqual = await this.compareFiles(
+                                file,
+                                this.configFileStructure.assets[this.lang].file(newAssetName),
+                                newAssetName
+                            );
+                            if (filesEqual) break;
+                        }
+                        newAssetName = `${file.name.split('.').at(0)}(${i}).${file.name.split('.').at(-1)}`;
+                        i++;
+                    }
+                }
+            }
+            uploadSource = `${this.configFileStructure.uuid}/assets/${this.lang}/${newAssetName}`;
+            this.configFileStructure.assets[this.lang].file(newAssetName, file);
+        }
+
+        if (this.sourceCounts[uploadSource]) {
+            this.sourceCounts[uploadSource] += 1 + oppositeSourceCount;
+        } else {
+            this.sourceCounts[uploadSource] = 1 + oppositeSourceCount;
+        }
+
+        let imageSrc = URL.createObjectURL(file);
+        return {
+            id: uploadSource,
+            altText: '',
+            caption: '',
+            src: imageSrc
+        };
     }
 
     onFileChange(e: Event): void {
         // create object URL(s) to display image(s)
         const filelist = Array.from((e.target as HTMLInputElement).files as ArrayLike<File>);
-        this.imagePreviews.push(
-            ...filelist.map((file: File) => {
-                // Add the uploaded images to the product ZIP file.
-                const uploadSource = `${this.configFileStructure.uuid}/assets/${this.lang}/${file.name}`;
-                this.configFileStructure.assets[this.lang].file(file.name, file);
-
-                if (this.sourceCounts[uploadSource]) {
-                    this.sourceCounts[uploadSource] += 1;
-                } else {
-                    this.sourceCounts[uploadSource] = 1;
-                }
-
-                let imageSrc = URL.createObjectURL(file);
-                return {
-                    id: file.name,
-                    altText: '',
-                    caption: '',
-                    src: imageSrc
-                };
-            })
-        );
-        this.onImagesEdited();
+        const filePromises = filelist.map((file: File) => this.addUploadedFile(file));
+        Promise.all(filePromises).then((files) => {
+            this.imagePreviews.push(...files);
+            this.onImagesEdited();
+        });
     }
 
     dropImages(e: DragEvent): void {
@@ -285,41 +450,27 @@ export default class ImageEditorV extends Vue {
                 files = [files[0]];
             }
 
-            this.imagePreviews.push(
-                ...files.map((file: File) => {
-                    // Add the uploaded images to the product ZIP file.
-                    const uploadSource = `${this.configFileStructure.uuid}/assets/${this.lang}/${file.name}`;
-                    this.configFileStructure.assets[this.lang].file(file.name, file);
+            const filePromises = files.map((file: File) => this.addUploadedFile(file));
+            Promise.all(filePromises).then((files) => {
+                this.imagePreviews.push(...files);
+                this.onImagesEdited();
+            });
 
-                    if (this.sourceCounts[uploadSource]) {
-                        this.sourceCounts[uploadSource] += 1;
-                    } else {
-                        this.sourceCounts[uploadSource] = 1;
-                    }
-
-                    let imageSrc = URL.createObjectURL(file);
-                    return {
-                        id: file.name,
-                        altText: '',
-                        caption: '',
-                        src: imageSrc
-                    };
-                })
-            );
             this.dragging = false;
         }
-        this.onImagesEdited();
     }
 
     deleteImage(img: ImageFile): void {
         const idx = this.imagePreviews.findIndex((file: ImageFile) => file.id === img.id);
         if (idx !== -1) {
-            const fileSource = `${this.configFileStructure.uuid}/assets/${this.lang}/${this.imagePreviews[idx].id}`;
+            const assetFolder = this.imagePreviews[idx].id.split('/')[2];
+            const assetSource = this.imagePreviews[idx].id;
+            const assetRelativePath = this.imagePreviews[idx].id.split('/').slice(3).join('/');
 
             // Remove the image from the product ZIP file.
-            this.sourceCounts[fileSource] -= 1;
-            if (this.sourceCounts[fileSource] === 0) {
-                this.configFileStructure.assets[this.lang].remove(this.imagePreviews[idx].id);
+            this.sourceCounts[assetSource] -= 1;
+            if (this.sourceCounts[assetSource] === 0) {
+                this.configFileStructure.assets[assetFolder].remove(assetRelativePath);
                 URL.revokeObjectURL(this.imagePreviews[idx].src);
             }
             this.imagePreviews.splice(idx, 1);
@@ -355,18 +506,20 @@ export default class ImageEditorV extends Vue {
                     // @ts-ignore
                     (this.panel as ImagePanel)[key] = imageFile[key];
                 });
-
-                (this.panel as ImagePanel).src = `${this.configFileStructure.uuid}/assets/${this.lang}/${imageFile.id}`;
+                (this.panel as ImagePanel).src = imageFile.id;
             } else {
                 // Otherwise, convert this to a slideshow panel.
                 this.panel.type = PanelType.Slideshow;
                 this.panel.caption = this.slideshowCaption ?? undefined;
 
-                // Turn each of the image configs into an image panel and add them to the slidesow.
+                // Turn each of the image configs into an image panel and add them to the slideshow.
                 (this.panel as SlideshowPanel).items = this.imagePreviews.map((imageFile: ImageFile) => {
+                    const imageSrc = imageFile.id;
+                    const imageId = imageFile.id.split('/').at(-1);
                     return {
                         ...imageFile,
-                        src: `${this.configFileStructure.uuid}/assets/${this.lang}/${imageFile.id}`,
+                        src: imageSrc,
+                        id: imageId,
                         type: PanelType.Image
                     } as ImagePanel;
                 });
--- a/src/components/metadata-editor.vue
+++ b/src/components/metadata-editor.vue
@@ -499,10 +499,13 @@
                 :configLang="configLang"
                 :saving="saving"
                 :unsavedChanges="unsavedChanges"
+                @shared-asset="updateToSharedAsset"
+                @process-panel="panelHelper"
                 @save-changes="onSave"
                 @save-status="updateSaveStatus"
                 @refresh-config="refreshConfig"
                 @export-product="exportProduct"
+                @lang-change="changeLang"
                 ref="mainEditor"
             >
                 <!-- Metadata editing modal inside the editor -->
@@ -745,7 +748,7 @@ export default class MetadataEditorV extends Vue {
     };
     slides: MultiLanguageSlide[] = [];
     sourceCounts: SourceCounts = {};
-    sessionExpired: boolean = false;
+    sessionExpired = false;
     totalTime = import.meta.env.VITE_APP_CURR_ENV ? Number(import.meta.env.VITE_SESSION_END) : 30;
 
     mounted(): void {
@@ -776,7 +779,6 @@ export default class MetadataEditorV extends Vue {
             this.metadata.tocOrientation = 'vertical';
             this.metadata.returnTop = true;
         }
-
         // Find which view to render based on route
         if (this.$route.name === 'editor') {
             this.loadEditor = true;
@@ -896,12 +898,12 @@ export default class MetadataEditorV extends Vue {
         this.handleSessionTimeout();
     }
 
-    processAsset(asset: string | undefined, name: string): Promise<any> {
-        if (!asset) return Promise.resolve();
+    processAsset(assetPath: string | undefined, name: string): Promise<any> {
+        if (!assetPath) return Promise.resolve();
 
         // Load asset (if provided).
         return new Promise((res) => {
-            const assetSrc = `assets/${this.configLang}/${name}`;
+            const assetSrc = assetPath.split('/').slice(1).join('/'); // Omit product UUID
             const assetFile = this.configFileStructure?.zip.file(assetSrc);
             const assetType = assetSrc.split('.').at(-1);
 
@@ -928,12 +930,12 @@ export default class MetadataEditorV extends Vue {
                 }
             } else {
                 // If it doesn't exist, maybe it's a remote file?
-                fetch(asset).then((data: Response) => {
+                fetch(assetPath).then((data: Response) => {
                     if (data.status !== 404) {
                         data.blob().then((blob: Blob) => {
                             res({
                                 file: new File([blob], name),
-                                preview: asset,
+                                preview: assetPath,
                                 external: true // indicates that this is an external asset
                             });
                         });
@@ -943,9 +945,13 @@ export default class MetadataEditorV extends Vue {
         });
     }
 
+    changeLang(lang: string): void {
+        this.configLang = lang;
+    }
+
     /**
      * Open current editor config as a new Storylines product in new tab.
-     * Note: Preview button on metadata editor will only show when editing an existing product, not cwhen creating a new one
+     * Note: Preview button on metadata editor will only show when editing an existing product, not when creating a new one
      * This is a design decision, can change if we decide that people would want to preview new products for some reason
      */
     preview(): void {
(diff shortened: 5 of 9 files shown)


RyanCoulsonCA (changes requested):
Haven't tested the demo for this PR just yet as I'm a bit tied up with TCEI changes, but this is my first-pass review just taking a look at the code. Will test the demo tomorrow!

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OAU191M8_oUQk6xXKjs:b19gumc)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @IshavSohal)

___
*[`src/components/image-editor.vue` line 327 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATy04KB2yi5KW_Iy2E:-OATy04KB2yi5KW_Iy2F:b-3gyugi) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L327)):*
> this.configFileStructure.assets['shared'].file(this.imagePreviews[idx].id) ? 'shared' : this.lang

This is used both here and in the `const fileSource` line assignment above, it may be worth assigning it to it's own variable at the top of the code block to prevent code duplication.
___
*[`src/components/image-editor.vue` line 351 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATwSqh7VNoEyHV3fPl:-OATwSqh7VNoEyHV3fPm:b-x3wn9f) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L351)):*
> ```Vue
>             console.log('imagePreviews');
>             console.log(JSON.parse(JSON.stringify(this.imagePreviews)));
>             for (let i = 0; i < this.panel.items.length; i++) {
> ```

If we don't need to use the item index for anything, I'm thinking we can just use the following to save a couple of lines:
```js 
for(asset in this.panel.items) {
  const assetSrc = asset.src.split('/');
  ...
}
```
___
*[`src/components/image-editor.vue` line 358 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATwxV72yMtAOK0ug5m:-OATwxV72yMtAOK0ug5n:b-45eoyo) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L358)):*
> ```Vue
>                 const assetFolder = assetSrc[2];
>                 const assetId = assetSrc[3];
>                 if (this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared') {
> ```

You can just use the following here:

```js
return this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared';
```
___
*[`src/components/image-editor.vue` line 372 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATx8MF1wg1auxTCo9q:-OATx8MF1wg1auxTCo9r:ba4o3tc) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L372)):*
> ```Vue
>             const assetFolder = assetSrc[2];
>             const assetId = assetSrc[3];
>             if (this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared') {
> ```

Same as above, you can just use this here:
```js
return this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared'
```
___
*[`src/components/image-editor.vue` line 425 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATyzQE6UwfQqOad0Gm:-OATyzQE6UwfQqOad0Gn:b7o1xqb) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L425)):*
> this.configFileStructure.assets['shared'].file(imageFile.id) ? 'shared' : this.lang

Same as before, since this is used multiple times (here and in the `else` block below), it may be worth assigning it earlier to reduce code duplication.
___
*[`src/components/metadata-editor.vue` line 879 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU-i85-mP-725X_XzH:-OAU-i85-mP-725X_XzI:b-kafufu) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/metadata-editor.vue#L879)):*
> ```Vue
> 
>         // If uploadLogo is set, upload the logo to the directory.
>         // Q: if we create a shared assets folder, should the logo just go there, since both langs
> ```

Someone can shut me down if I'm wrong here, but I don't know if we should necessarily assume that the logo will be the same for both products. The best example I can think of would be if the product logo has some sort of text in it and the author wants to have an English and a French version of it.

I'm not sure how common of a case that would be though. It may be worth a discussion at some point.
___
*[`src/components/video-editor.vue` line 333 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU1UtV-7VomXHqiBsF:-OAU1UtV-7VomXHqiBsG:bhq4njz) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/video-editor.vue#L333)):*
> this.configFileStructure.assets['shared'].file(this.videoPreview.id) ? 'shared' : this.lang

Same as the image file, this should probably be stored assigned to a variable earlier to prevent code duplication.
___
*[`src/components/video-editor.vue` line 377 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU1hU64UdpxbMf0VpK:-OAU1hU64UdpxbMf0VpL:bjr34lz) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/video-editor.vue#L377)):*
> ```Vue
>             // An assets src value is "out of date" if the asset exists in the shared asset folder, but still
>             // references the curr lang's assets folder
>             if (this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared') {
> ```

```js
return this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared'
```
___
*[`src/components/video-editor.vue` line 407 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU1ox13QCQPXDUo_vt:-OAU1ox13QCQPXDUo_vu:b-2e9313) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/video-editor.vue#L407)):*
> this.configFileStructure.assets['shared'].file(this.videoPreview.id) ? 'shared' : this.lang

Same as the previous comments about this!

szczz (changes requested):
>When deleting an asset and its source count hits 0, it gets removed from `configFileStructure`. However the asset doesn't get removed from the assets folder of the product. Due to this, when the app is reloaded and `configFileStructure` is initialized, the deleted asset returns. If this is a bug, I can open a new issue for this.

This seems like a bug to me. We don't want stale assets floating around in storylines products. Someone please correct me if I'm wrong and missing a use case for this. 

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OAYYm-wBAL_rlJRbzCB:bzidcz7)* status: all files reviewed, 10 unresolved discussions (waiting on @IshavSohal and @RyanCoulsonCA)

___
*[`src/components/image-editor.vue` line 225 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAYXTf8-7KiKrqwJBq0:-OAYXTf8-7KiKrqwJBq1:b2orb1x) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L225)):*
> ```Vue
>             oppositeSourceCount = this.sourceCounts[oppositeFileSource] ?? 0;
>             this.sourceCounts[oppositeFileSource] = 0;
>             this.configFileStructure.assets[oppositeLang].remove(file.name);
> ```

I think we need to perform this on this.lang's folder. While following the PR test steps, I noticed that the img remains in the original langs asset folder after it has been moved to the shared folder. 

Scenario "Adding an asset to configs of both langs" indicates that the image should be removed from the assets/en folder, however, it persists. 
___
*[`src/components/metadata-editor.vue` line 879 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU-i85-mP-725X_XzH:-OAYYSQu4uyI75VpggCn:bzahgee) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/metadata-editor.vue#L879)):*

Someone can shut me down if I'm wrong here, but I don't know if we should necessarily assume that the logo will be the same for both products. The best example I can think of would be if the product logo has some sort of text in it and the author wants to have an English and a French version of it.

I'm not sure how common of a case that would be though. It may be worth a discussion at some point.
</blockquote></details>

Agreed and I think that will be a relatively common case.

RyanCoulsonCA (changes requested):
> When deleting an asset and its source count hits 0, it gets removed from `configFileStructure`. However the asset doesn't get removed from the assets folder of the product. Due to this, when the app is reloaded and `configFileStructure` is initialized, the deleted asset returns. If this is a bug, I can open a new issue for this.

If I'm understanding this correctly, it's probably this same issue we've had for a while. See the following: https://github.com/ramp4-pcar4/storylines-editor/issues/343, https://github.com/ramp4-pcar4/storylines-editor/issues/111

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OAZ1buw4uLSIpWK8Qsh:brvktdq)* status: all files reviewed, 11 unresolved discussions (waiting on @IshavSohal)

___
*[`src/components/image-editor.vue` line 221 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAYx8Cf545Nt5fPeJG9:-OAYx8Cf545Nt5fPeJGA:bcokl9i) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L221)):*
> ```Vue
>         // in the shared nor the opposite langs assets folder, the file should be uploaded to the current
>         // langs assets folder
>         if (this.configFileStructure.assets[oppositeLang].file(file.name)) {
> ```

I've found a bug that most likely relates to this section. Try out these steps:

1. Create a new image panel using the English configuration.
2. Upload an image.
3. Save the product.
4. Swap to the French configuration and create a new image panel.
5. Upload the same image you uploaded to the English slide.
6. Save the product.

The image is correctly moved to the `shared` folder, however, the image source for the panel in the English configuration still references the `en` folder when it should be referencing the `shared` folder.

This block will need additional code that updates the source path in the opposite language config with the new `shared` path. It looks like the video editor is missing this too, so we'll need there too.

RyanCoulsonCA (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: all files reviewed, 11 unresolved discussions (waiting on @IshavSohal)

___
*[`src/components/image-editor.vue` line 351 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATwSqh7VNoEyHV3fPl:-OAdNvCL1i7I0cDYlTsJ:b-uyzha6) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L351)):*

If we don't need to use the item index for anything, I'm thinking we can just use the following to save a couple of lines:
```js 
for(asset in this.panel.items) {
  const assetSrc = asset.src.split('/');
  ...
}
```
</blockquote></details>

In hindsight, I don't think it actually works as I've written it here. You can ignore it, or `this.panel.items.forEach(asset => { ... });` should work.

IshavSohal (commented):
Yeah that's exactly what I was referring to

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OAsRt2D9GWy8rNbTWgK:b-j1ives)* status: 2 of 7 files reviewed, 11 unresolved discussions (waiting on @RyanCoulsonCA and @szczz)

___
*[`src/components/image-editor.vue` line 221 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAYx8Cf545Nt5fPeJG9:-OAsS3dkBOBO5xE24dPJ:b-l4dcf9) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L221)):*

I've found a bug that most likely relates to this section. Try out these steps:

1. Create a new image panel using the English configuration.
2. Upload an image.
3. Save the product.
4. Swap to the French configuration and create a new image panel.
5. Upload the same image you uploaded to the English slide.
6. Save the product.

The image is correctly moved to the `shared` folder, however, the image source for the panel in the English configuration still references the `en` folder when it should be referencing the `shared` folder.

This block will need additional code that updates the source path in the opposite language config with the new `shared` path. It looks like the video editor is missing this too, so we'll need there too.
</blockquote></details>

I ended up creating a `shared-asset` event, which is emitted whenever an asset is to be moved to the shared assets folder. The handler for this event will look for each instance of the uploaded asset within the opposite lang's config, and will update its source to refer to the shared asset folder. Is this the approach you had in mind? Let me know if you still have this bug. 
___
*[`src/components/image-editor.vue` line 225 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAYXTf8-7KiKrqwJBq0:-OAsSUgF9Vx2oUbP--nu:b-n1l0dm) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L225)):*
<details><summary><i>Previously, szczz (Matt Szczerba) wrote…</i></summary><blockquote>

I think we need to perform this on this.lang's folder. While following the PR test steps, I noticed that the img remains in the original langs asset folder after it has been moved to the shared folder. 

Scenario "Adding an asset to configs of both langs" indicates that the image should be removed from the assets/en folder, however, it persists.
</blockquote></details>

I think this may be related to the bug I mentioned (the assets not being deleted from the product in the server). Whenever the product is re-fetched from the server, `configFileStructure` gets updated such that it contains any/all deleted assets. 

If the product is not re-fetched from the server, this issue shouldn't occur, but correct me if I'm wrong.  
___
*[`src/components/image-editor.vue` line 327 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATy04KB2yi5KW_Iy2E:-OAsT1vk0fS2sZ6__A4t:b2tpwxa) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L327)):*

> this.configFileStructure.assets['shared'].file(this.imagePreviews[idx].id) ? 'shared' : this.lang

This is used both here and in the `const fileSource` line assignment above, it may be worth assigning it to it's own variable at the top of the code block to prevent code duplication.
</blockquote></details>

Donethanks, I created a method `srcFolder()` to do this
___
*[`src/components/image-editor.vue` line 351 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATwSqh7VNoEyHV3fPl:-OAsTAkrBszPZBqDYY9a:bikqny) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L351)):*

In hindsight, I don't think it actually works as I've written it here. You can ignore it, or `this.panel.items.forEach(asset => { ... });` should work.
</blockquote></details>

Due to the new `shared-asset` event this method isn't needed anymore, so I just deleted it. 
___
*[`src/components/image-editor.vue` line 358 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATwxV72yMtAOK0ug5m:-OAsTKrJ6-6xHUCnY2i6:b4tguxp) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L358)):*

You can just use the following here:

```js
return this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared';
```
</blockquote></details>

Deleted this method, all good now
___
*[`src/components/image-editor.vue` line 372 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATx8MF1wg1auxTCo9q:-OAsU2nZ16ZgJVNra1I1:b4tguxp) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L372)):*

Same as above, you can just use this here:
```js
return this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared'
```
</blockquote></details>

Deleted this method, all good now
___
*[`src/components/image-editor.vue` line 425 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATyzQE6UwfQqOad0Gm:-OAsUAVS8TARr5thT7Gw:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L425)):*

> this.configFileStructure.assets['shared'].file(imageFile.id) ? 'shared' : this.lang

Same as before, since this is used multiple times (here and in the `else` block below), it may be worth assigning it earlier to reduce code duplication.
</blockquote></details>

Donethanks
___
*[`src/components/metadata-editor.vue` line 879 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU-i85-mP-725X_XzH:-OAsTbSB2OxGXnuZgzfN:bo9s5pi) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/metadata-editor.vue#L879)):*
<details><summary><i>Previously, szczz (Matt Szczerba) wrote…</i></summary><blockquote>

Agreed and I think that will be a relatively common case.
</blockquote></details>

Ah fair point. However upon creating a product, both configs now have a `src` for their respective logos that refers to the `assets/en` folder. 
___
*[`src/components/video-editor.vue` line 333 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU1UtV-7VomXHqiBsF:-OAsUD414vVvqn2bELUl:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/video-editor.vue#L333)):*

> this.configFileStructure.assets['shared'].file(this.videoPreview.id) ? 'shared' : this.lang

Same as the image file, this should probably be stored assigned to a variable earlier to prevent code duplication.
</blockquote></details>

Donethanks
___
*[`src/components/video-editor.vue` line 377 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU1hU64UdpxbMf0VpK:-OAsUQ7YBevuBI8lk_Hk:b4tguxp) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/video-editor.vue#L377)):*

```js
return this.configFileStructure.assets['shared'].file(assetId) && assetFolder !== 'shared'
```
</blockquote></details>

Deleted this method, all good now
___
*[`src/components/video-editor.vue` line 407 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OAU1ox13QCQPXDUo_vt:-OAsUFlMB8DnElb95nqZ:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/video-editor.vue#L407)):*

> this.configFileStructure.assets['shared'].file(this.videoPreview.id) ? 'shared' : this.lang

Same as the previous comments about this!
</blockquote></details>

Donethanks

RyanCoulsonCA (changes requested):
> Reviewed 2 of 6 files at r1, 4 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 6 of 7 files reviewed, 4 unresolved discussions (waiting on @IshavSohal and @szczz)

___
*[`src/components/editor.vue` line 218 at r2](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCJgxq87yQ7dG1H2es1:-OCJgxq87yQ7dG1H2es2:b-aqarlv) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5844a963cdb3d1942f54cbcf35baa9e847cc03d4/src/components/editor.vue#L218)):*
> Is this the approach you had in mind?

I think this approach should work :thumbsup: 
___
*[`src/components/editor.vue` line 229 at r2](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCJhMv0C2cJEHtA5ULQ:-OCJhMv0C2cJEHtA5ULR:blb3f7v) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5844a963cdb3d1942f54cbcf35baa9e847cc03d4/src/components/editor.vue#L229)):*
> ```Vue
>         console.log(JSON.parse(JSON.stringify(oppositeConfig.slides)));
> 
>         for (let i = 0; i < oppositeConfig.slides.length; i++) {
> ```

This (and the `panel` loop just below this) can be simplified with something like the following:

```js
oppositeConfig.slides.forEach(oppositeSlide => { ... });
```

The variable `oppositeSlide` can be changed to whatever you think works!
___
*[`src/components/editor.vue` line 239 at r2](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCJhprGB7kDGK_goLCV:-OCJhprGB7kDGK_goLCW:b-3lkibf) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5844a963cdb3d1942f54cbcf35baa9e847cc03d4/src/components/editor.vue#L239)):*
> ```Vue
>                 console.log(JSON.parse(JSON.stringify(currPanel)));
> 
>                 // if panel is slideshow, image or video then proceed
> ```

This will miss the `dynamic` panel, which like slideshows can have any slide type as a child (including slideshows).

It may be nice to use a recursive solution here like we use in other parts of the app in order to reduce code duplication (see an example [here](https://github.com/ramp4-pcar4/storylines-editor/blob/main/src/components/metadata-editor.vue#L784), although we can simplify it). 

The recursive function would look something like the following (disclaimer: wrote this directly in Reviewable and haven't tested it so it may have some syntax errors or bugs):

```js
const recursiveHelper = (panel: BasePanel) => {
  switch (panel.type) {
    case 'slideshow':
      panel.items.forEach(item => recursiveHelper(item));
      break;
    case 'dynamic':
      panel.children.forEach(child => recursiveHelper(child));
      break;
    default:
      // your code to handle the individual panels here.
      // If we want to ensure this only runs on images and videos, we can add a check for that here too.
      break;
  }
}
```

After you have this, you should just need to call it like this:
```js
currSlide.panel.forEach(panel => { recursiveHelper(panel); }
```

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 2 of 9 files reviewed, 5 unresolved discussions (waiting on @RyanCoulsonCA and @szczz)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCeC0kT8e76WOlaSLJr:-OCeC0kT8e76WOlaSLJs:b-jbw7is) (no related file):*
I've made one additional change. With the new redesigns merged in,  I noticed that the `configLang` prop was not updating appropriately when changing from the `en` config to the `fr` config (since they now both reside in the same editor view). I've created a new `lang-change` event that is emitted whenever a `slide-change` event is handled by the `editor` component. The `lang-change` event is handled in the `metadata-editor` component. 
___
*[`src/components/editor.vue` line 229 at r2](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCJhMv0C2cJEHtA5ULQ:-OCeBvqk2HwHHHBXd-9t:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5844a963cdb3d1942f54cbcf35baa9e847cc03d4/src/components/editor.vue#L229)):*

This (and the `panel` loop just below this) can be simplified with something like the following:

```js
oppositeConfig.slides.forEach(oppositeSlide => { ... });
```

The variable `oppositeSlide` can be changed to whatever you think works!
</blockquote></details>

Donethanks
___
*[`src/components/editor.vue` line 239 at r2](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCJhprGB7kDGK_goLCV:-OCeBxKbC67QIlEYLhWu:b-5a2h77) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5844a963cdb3d1942f54cbcf35baa9e847cc03d4/src/components/editor.vue#L239)):*

This will miss the `dynamic` panel, which like slideshows can have any slide type as a child (including slideshows).

It may be nice to use a recursive solution here like we use in other parts of the app in order to reduce code duplication (see an example [here](https://github.com/ramp4-pcar4/storylines-editor/blob/main/src/components/metadata-editor.vue#L784), although we can simplify it). 

The recursive function would look something like the following (disclaimer: wrote this directly in Reviewable and haven't tested it so it may have some syntax errors or bugs):

```js
const recursiveHelper = (panel: BasePanel) => {
  switch (panel.type) {
    case 'slideshow':
      panel.items.forEach(item => recursiveHelper(item));
      break;
    case 'dynamic':
      panel.children.forEach(child => recursiveHelper(child));
      break;
    default:
      // your code to handle the individual panels here.
      // If we want to ensure this only runs on images and videos, we can add a check for that here too.
      break;
  }
}
```

After you have this, you should just need to call it like this:
```js
currSlide.panel.forEach(panel => { recursiveHelper(panel); }
```
</blockquote></details>

Donethanks, appreciate the help!

RyanCoulsonCA (changes requested):
> Reviewed 5 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @IshavSohal and @szczz)

___
*[`server/index.js` line 12 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-ODHw-Mt3II87xnCocAk:-ODHw-Mt3II87xnCocAl:baa49t0) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/0006581cbe51b81dd6e1f45b9df2dbf4a9fb1d31/server/index.js#L12)):*
> ```JavaScript
> const simpleGit = require('simple-git');
> const responseMessages = [];
> require('dotenv').config();
> ```

I think this file snuck into this PR somehow! Other than that, all of the test I've done look good!

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 5 of 9 files reviewed, 3 unresolved discussions (waiting on @RyanCoulsonCA and @szczz)

___
*[`server/index.js` line 12 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-ODHw-Mt3II87xnCocAk:-ODmgyTxAvPGnekeGASA:b-rhtruj) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/0006581cbe51b81dd6e1f45b9df2dbf4a9fb1d31/server/index.js#L12)):*

I think this file snuck into this PR somehow! Other than that, all of the test I've done look good!
</blockquote></details>

This was just me fixing some formatting issues I noticed. Everything should be working the same as before.

RyanCoulsonCA (approved):
> Reviewed 1 of 7 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: all files reviewed, 3 unresolved discussions (waiting on @IshavSohal and @szczz)

___
*[`server/index.js` line 12 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-ODHw-Mt3II87xnCocAk:-OE-sg0_3yiyLgCgqYq2:b4c9lyr) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/0006581cbe51b81dd6e1f45b9df2dbf4a9fb1d31/server/index.js#L12)):*

This was just me fixing some formatting issues I noticed. Everything should be working the same as before.
</blockquote></details>

:thumbsup:

szczz (changes requested):
#### Adding an asset to configs of both langs

After performing these steps, but with a slight modification of copying the config from the en slide. The assets still remain in the en folder. They get moved to the shared folder if you manually add the same image to the FR slide though.

> Reviewed 1 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OE-z3_cASd6P4rEcAQo:bmep66x)* status: all files reviewed, 2 unresolved discussions (waiting on @IshavSohal and @RyanCoulsonCA)

___
*[`server/index.js` line 12 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-ODHw-Mt3II87xnCocAk:-ODlstpz65BA8qXe6jd2:bkgk8fy) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/0006581cbe51b81dd6e1f45b9df2dbf4a9fb1d31/server/index.js#L12)):*

I think this file snuck into this PR somehow! Other than that, all of the test I've done look good!
</blockquote></details>

Yep please remove this file from the PR

IshavSohal (commented):
Donethanks. As I was making the fix for this, I decided to create a helper method, `panelHelper()`, which can be provided a callback (as well as additional arguments for it) to be called on each ImagePanel/VideoPanel within a given BasePanel. This may be a bit overkill, so let me know your thoughts/opinions. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OEpR4kJE9BjpH8HfDqv:b-gtmzkt)* status: 4 of 10 files reviewed, 1 unresolved discussion (waiting on @RyanCoulsonCA and @szczz)

___
*[`server/index.js` line 12 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-ODHw-Mt3II87xnCocAk:-OEpR1bz-cwyPm1WvZgs:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/0006581cbe51b81dd6e1f45b9df2dbf4a9fb1d31/server/index.js#L12)):*
<details><summary><i>Previously, szczz (Matt Szczerba) wrote…</i></summary><blockquote>

Yep please remove this file from the PR
</blockquote></details>

Donethanks

RyanCoulsonCA (changes requested):
> Reviewed 4 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 8 of 10 files reviewed, 1 unresolved discussion (waiting on @IshavSohal and @szczz)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCeC0kT8e76WOlaSLJr:-OG_xyxP6JAw4i-b8Kjr:b-6njivp) (no related file):*

I've made one additional change. With the new redesigns merged in,  I noticed that the `configLang` prop was not updating appropriately when changing from the `en` config to the `fr` config (since they now both reside in the same editor view). I've created a new `lang-change` event that is emitted whenever a `slide-change` event is handled by the `editor` component. The `lang-change` event is handled in the `metadata-editor` component.
</blockquote></details>

Found a bug:

1. Create a new blank slide.
2. Change both the English and French slide to an "image" slide.
3. Upload a photo to the English slide.
4. Upload the same photo to the French slide.
5. Switch back to the English slide and notice that the image doesn't appear. Switching over to the Advanced panel shows that the `src` still points to the `en` folder instead of `shared`.

This seems to work fine for slideshow panels, just breaks with image panels.

yileifeng (commented):
Note: loading in existing products with old configuration asset folder structure will freeze the metadata page and eventually resolve with all N/A values. This breaks backwards compatibility and should be considered before merging without a plan in place to migrate old products.

> Reviewed 1 of 6 files at r1, 2 of 7 files at r3, 1 of 3 files at r4, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OGa5dAADvMaBz3BVIx0:bq3d0g3)* status: all files reviewed, 2 unresolved discussions (waiting on @IshavSohal)

___
*[`src/components/image-editor.vue` line 130 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OATn-er4wesariRl3DN:-OATn-er4wesariRl3DO:b-2dr7hi) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/4e47d8b5f3847d233377a64056299aa9d33d6c68/src/components/image-editor.vue#L130)):*
> ```Vue
> 
>     mounted(): void {
>         console.log('');
> ```

Blocker for console logs

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @RyanCoulsonCA and @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OCeC0kT8e76WOlaSLJr:-OGacRBf7h5McoMQM-xS:b-d8zw1m) (no related file):*

Found a bug:

1. Create a new blank slide.
2. Change both the English and French slide to an "image" slide.
3. Upload a photo to the English slide.
4. Upload the same photo to the French slide.
5. Switch back to the English slide and notice that the image doesn't appear. Switching over to the Advanced panel shows that the `src` still points to the `en` folder instead of `shared`.

This seems to work fine for slideshow panels, just breaks with image panels.
</blockquote></details>

Donethanks, should be all good now

IshavSohal (commented):
This may have been due to the console logs that I've now removed, which were problematic for large products.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OGlElZl-XGm9K0g1pva:b-fefdeu)* status: 1 of 10 files reviewed, 3 unresolved discussions (waiting on @RyanCoulsonCA, @szczz, and @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OGlEjk55tWqE1N9Ct2U:-OGlEjk65SiJVhpR5Jqo:b1vwyqa) (no related file):*
Removed all console logs.

szczz (changes requested):
Couple of issues: 

1. If you create a new storyline, upload a logo to the EN config and then save it, the logo gets added to the en assets folder as expected. If you return to the dashboard and load the storyline and then change to the french config, the original logo remains. If you go to the editor page and then press save, it results in the logo being stored in both the en fold and fr folder even though its exactly the same.

2. If you create a new slide and upload an image with the name "test.jpeg" in the en config, then flip to the fr config and upload a different image with the name "test.jpeg", returning to the en config will display the second image. If you save the en config before the french one, the second image will show up in the shared folder, but the first image will remain in the en folder. Perhaps a hash function or looking at image metadata could help handle this situation. 
 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OGow2TZ4G-Ni7rH8NDN:b885t7t)* status: 1 of 10 files reviewed, 3 unresolved discussions (waiting on @IshavSohal, @RyanCoulsonCA, and @yileifeng)

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @RyanCoulsonCA, @szczz, and @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OHJb-zf8vhuBSn0HEnR:-OHJb-zf8vhuBSn0HEnS:b-osjsq4) (no related file):*
I have a few questions regarding how to proceed: 

When should we consider two asset files to be the same? Should it be when both their names and contents are the same, or when only their contents are the same? Currently I compare using both names and contents, but I'm running into some issues.

Should we expect assets to be in nested subfolders within the asset folders (like in the 0000... product)? Currently I'm doing this, but not sure if this is unnecessary.

IshavSohal:

Comparing file content to determine equality seems reasonable IMO. However, what would the naming convention of the shared asset file be in this situation?
</blockquote></details>

There are a few edge cases when comparing just file contents:

- Suppose there exists an asset, `test.png` in the `en` folder
- If you try to upload an asset `test2.png` with the same contents as `test.png` into the `en` config, then `test.png` would be used as the src, meaning that `test2.png` would not be uploaded to the `en` folder
- If you try to upload an asset `test.png` with different contents than the original `test.png` into the `en` config, we would have to upload this asset into the `en` folder using a unique name, like `test(2).png`

When comparing both contents and names, however, there are many more edge cases. Are there any use cases in which the user would purposely want two files with the same content but different names?

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 7 of 13 files reviewed, 4 unresolved discussions (waiting on @RyanCoulsonCA, @szczz, and @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OHxIv7w8aDMHMgCUR8a:-OHxIv7w8aDMHMgCUR8b:b-c1a0qf) (no related file):*
There seems to be quite a bit of repetition regarding the code responsible for comparing files. Would it be a good idea to encapsulate these methods in a plugin? If so, should we wait until we migrate over to the Composition API?

yileifeng (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: all files reviewed, 4 unresolved discussions (waiting on @IshavSohal, @RyanCoulsonCA, and @szczz)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OHxIv7w8aDMHMgCUR8a:-OHxldBRCOOOBopwhcOB:b-fj4a61) (no related file):*

Will always advocate for prioritizing the code refactor/composition API migration. Maybe a patch solution for this PR (if not too difficult to implement) is to add a composable: https://vuejs.org/guide/reusability/composables.
</blockquote></details>

As a follow up, it does not make much sense to implement the composables approach if it will just be overwritten later. I would say it's fine to hold off on that implementation for now but it can be worth considering as an option in the refactor. It makes much more sense to merge this PR before starting on composition API / Pinia work.

IshavSohal (commented):
Now, when a new product is created, the logo is uploaded into the `shared` asset folder, since the config of the current lang is copied over to the opposite lang's config. However, if the user wants to change the logo for a specific lang's config, the asset uploaded will be added into that lang's asset folder (unless it already exists in the opposite lang's asset folder, in which case it gets uploaded to the `shared` folder)

Also, when the asset uploaded has the same name as an existing asset, the hashes of the two assets will be compared to ensure we are not uploading the same asset again. If the two assets have different content, however, the uploaded asset will be assigned a unique name to prevent overwriting the existing asset. As @yileifeng mentioned, it may be a good idea to allow the user to choose between creating their own unique name and overwriting the existing asset, so I'm going to try to get that working. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OICu_T376s83YpH4r0v:bmu77mu)* status: all files reviewed, 4 unresolved discussions (waiting on @RyanCoulsonCA, @szczz, and @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OHJb-zf8vhuBSn0HEnR:-OICrpFK569o3XUWcXQw:b-uz7krd) (no related file):*

This PR is growing increasingly large as we attempt to handle a fairly rare case which is a mild concern, we can always spinoff issues that refine niche cases for the feature. Thinking about it more, this would be my opinion for the following cases: 
1. Different file names - *always upload new file to assets folder* since it could be unexpected behavior for users to find their new file `test2.png` to not be added

This way we can also save a ton of computation time as we would only need to compare file contents when a newly uploaded file contains the same name.  I'm not really sure why this would be an actual use case but maybe if user wanted to upload the same image to different sections of the Storyline with different labels for clarity (e.g. intro_logo.png and slide3_logo.png). 

2. Same file names - this is the case when we compare file contents to avoid duplication and also addresses Matt's found bug

In this case, when the file contents differ an option is to prompt the users to rename or overwrite(?) the new file. Let me know if this overall approach makes sense or what might not work.
</blockquote></details>

Yeah this is how I'm currently doing it. I only compare the hashes of two asset files when they have the same name. If the asset uploaded shares a name with no other asset in any asset folder, it gets uploaded right away, without hashing any files. However, the name of the uploaded image would have to be compared with the names of all assets in each of the assets folder in this case, which could be time consuming. I'm not sure if we can really avoid this though.

Prompting the user to rename the uploaded asset seems like an interesting idea. Currently when the asset uploaded has the same name but different content as an existing asset, the asset uploaded gets a number (ex. `test(2).png`) appended to its name to ensure it's unique. This could potentially confuse users. I'll try giving this a shot.

Also, once the repetitive code is moved into its own plugin/composable, the size of this PR will reduce by ~300-400 lines.

IshavSohal:

Yeah this is how I'm currently doing it. I only compare the hashes of two asset files when they have the same name. If the asset uploaded shares a name with no other asset in any asset folder, it gets uploaded right away, without hashing any files. However, the name of the uploaded image would have to be compared with the names of all assets in each of the assets folder in this case, which could be time consuming. I'm not sure if we can really avoid this though.

Prompting the user to rename the uploaded asset seems like an interesting idea. Currently when the asset uploaded has the same name but different content as an existing asset, the asset uploaded gets a number (ex. `test(2).png`) appended to its name to ensure it's unique. This could potentially confuse users. I'll try giving this a shot.

Also, once the repetitive code is moved into its own plugin/composable, the size of this PR will reduce by ~300-400 lines.
</blockquote></details>

On second thought, I think it may be a better idea to save the additional feature (prompting user for new asset name) for another time, as it makes the code even more complicated. I think the current convention (adding a number to end of asset name) should be fine for the time being. 

I'll probably log a new issue once this PR is finalized and merged.

yileifeng (approved):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 7 of 13 files reviewed, 5 unresolved discussions (waiting on @IshavSohal, @RyanCoulsonCA, and @szczz)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OHJb-zf8vhuBSn0HEnR:-OIQo1sk-i6Tzvy0TDwW:b-fox8nq) (no related file):*

On second thought, I think it may be a better idea to save the additional feature (prompting user for new asset name) for another time, as it makes the code even more complicated. I think the current convention (adding a number to end of asset name) should be fine for the time being. 

I'll probably log a new issue once this PR is finalized and merged.
</blockquote></details>

All cases seem to be working well when testing :+1:
___
*[`src/components/image-editor.vue` line 295 at r12](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OIQoFP--gZXuwKMeAer:-OIQoFP--gZXuwKMeAes:bdfybr7) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/a3cf5507d87bd5741299fc0560ebd41dd9bd9ee6/src/components/image-editor.vue#L295)):*
> ```Vue
>             // Only compare the hashes of two files when they have the exact same name
>             if (assetName === file.name) {
>                 console.log('current image has same name as uploaded image');
> ```

Reminder to cleanup console logs all before merging

szczz (approved):
:trophy: 

> Reviewed 6 of 6 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-:-OIlDjTwDaPdm2fd7zMe:b-fnwc0l)* status: all files reviewed, 4 unresolved discussions (waiting on @IshavSohal and @RyanCoulsonCA)

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 8 of 13 files reviewed, 5 unresolved discussions (waiting on @RyanCoulsonCA, @szczz, and @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OIwBOJl4sGJuOi45Gqq:-OIwBOJl4sGJuOi45Gqr:bdorymn) (no related file):*
Removed all comments. Also integrated the new background image feature with the shared asset system. It seems to be working well on my end, but maybe try testing that out before merging.

yileifeng (approved):
> Reviewed 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: all files reviewed, 4 unresolved discussions (waiting on @IshavSohal and @RyanCoulsonCA)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OIwBOJl4sGJuOi45Gqq:-OJ-aEcq1H5IJpHeY6qz:bnfp4nl) (no related file):*

Removed all comments. Also integrated the new background image feature with the shared asset system. It seems to be working well on my end, but maybe try testing that out before merging.
</blockquote></details>

:lgtm:

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: 8 of 13 files reviewed, 7 unresolved discussions (waiting on @mohsin-r, @RyanCoulsonCA, and @yileifeng)

___
*[`src/components/editor.vue` line 537 at r13](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OJJUXTGBxDp6uXR_Aky:-OJPgu0HBGdBYmahW_Jt:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/34cc56612981fde3e38499668d30d9e28068c681/src/components/editor.vue#L537)):*

This line can be shortened to `const newLang = lang || this.configLang || 'en'`
</blockquote></details>

Donethanks
___
*[`src/components/image-editor.vue` line 285 at r13](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414#-OJJVO1L56wno_AI2w8o:-OJPgv6QB_0FPWc28cH-:behjw3g) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/34cc56612981fde3e38499668d30d9e28068c681/src/components/image-editor.vue#L285)):*

Curious which part of this method is `async`. Don't see any `await`s below unless there's one missing.
</blockquote></details>

Good catch, donethanks

mohsin-r (approved):
> Reviewed 5 of 5 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/414)* status: all files reviewed, 5 unresolved discussions (waiting on @IshavSohal and @RyanCoulsonCA)

Issue #419, opened 2024-11-01
Change UUID doesn't work after entering an existing UUID


IshavSohal:
## Description

When changing the UUID of a product, if you first enter an existing UUID , then enter a unique UUID , the unique UUID gets switched back to the products UUID. 

## Steps to reproduce

1. Load in an existing product
2. Click on `Change UUID`
3. Enter an existing UUID (ex. the same as the current product)
4. Observe the `UUID already in use` alert
5. Change the UUID to a unique one, and observe the alert disappearing 
6. Press `Rename`
7. Observe that the UUID within the text input switches back to the products UUID, and that the alert references the products UUID
8. Click `Save changes` and observe that a product with the unique UUID you entered isn't created
9. Refresh the page, load an existing product, and click on  `Change UUID` again
10. This time enter a unique UUID to begin with
11. Press `Rename`
12. Observe that the UUID within the text input is the unique UUID you entered, and that the alert references the unique UUID
13. Click `Save changes` and observe that a product with the unique UUID you entered is created

## Expected behavior

After entering an existing UUID, you should still be able to enter a unique UUID and successfully change the products UUID

## Actual behavior

After entering an existing UUID, you are not able to enter a unique UUID and change the products UUID

Issue #418, opened 2024-10-31
Change uuid alert persists after saving changes


IshavSohal:
## Description

When changing the uuid of a product, the following alert appears: `You are changing this product UUID to <new uuid>. Save changes required`,. However when `save changes` is clicked, the alert persists

## Steps to reproduce

1. Load an existing product
2. Click the change uuid link
3. Enter a new uuid and click `rename`
4. Observe the alert
5. Click `save changes`
6. Observe that the alert persists

## Expected behavior

Alert should be removed upon saving

## Actual behavior

Alert persists

RyanCoulsonCA:
This will be fixed by https://github.com/ramp4-pcar4/storylines-editor/pull/487

Issue #253, opened 2024-02-29
Renaming Storylines products


dan-bowerman:
Worth a discussion:

- Can we provide capability to change a Storylines ID? 
- What would be involved? What checks should we do to ensure the renaming was successful? 
- Can a "find/replace" of an ID work through a Storylines JSON structure? 
- Where would we allow this in RESPECT?

szczz:
Adding some thoughts here as I think this need to be fleshed out a bit more. 

- There appears to be a race condition or an issue displaying the change UUID option when loading larger storylines
- After entering a new UUID and pressing Rename, a warning message appears saying "You are changing this product UUID to {original name}. Save changes required". It doesn't seem to be pulling the name from the previous input and just reverts back to the original name. 
- Instead of forcing a press of save changes we can use a modal or display the warning prior to the user pressing the Rename option. Reduces mouse clicks & confusion. 
- Renaming should use its own function and not generateConfig. Currently, renaming a storyline results in a new storyline being created instead of renaming the original. 

Additional notes:

- .NET API endpoint has been created to rename all instances of the old UUID with the new one. See the post request added to renameProduct on DevDeployment for details. 
- Putting the urgent tag on this as we've released this feature and should probably get it working correctly ASAP.

PR #487, opened 2024-12-09
re-implement rename feature

RyanCoulsonCA:
### Related Item(s)
#253 
#418 
#419

### Changes
- Added the `/rename` server endpoint: this is a re-implementation of the rename feature. Instead of relying on the `/upload` endpoint that simply created a new product folder, the `/rename` endpoint calls the `rename` syscall on the existing product folder. It will also delete and re-instantiate the Git repo.
- Added a new `/check` endpoint to the server to see if a product already exists. This way, we don't need to call the `/retrieve` endpoint and wait for the entire product to transfer every time the user enters a UUID. It should be much faster now.
- Moved the `responseMessages.push()` calls in to the `logger` function to reduce code duplication.

### Notes
image

### Testing
This is a big one, so please test thoroughly! If my instructions below are confusing please reach out, or post a comment on this PR to help others if they also have trouble.

**:warning: IMPORTANT! :warning:** If testing locally, please move the `server` folder outside of the `storylines-editor` folder. `npm run dev` seems to lock the `server/public` folder and it will result in the `rename` system call hanging indefinitely otherwise. Alternatively, if you overwrite the `server/index.js` code with the new code in this PR, you should be able to test using the demo page associated with this PR.<br />
**:warning:** If you notice the rename endpoint is hanging at any point, then the product folder is being accessed from somewhere else at the same time. Try closing anything that could be using it (VS Code, File Explorer, notepad, etc.). As far as I understand this should (hopefully) not happen on prod as nothing should be accessing the file system. If anyone knows of something that can be added to the code to fix this, please let me know.

Steps:
1. Pull this PR locally and follow the important step above.
2. Try creating a new product with an existing UUID. Notice that the "UUID is already in use" error should appear much faster than before.
3. Create a new product that doesn't already exist. Save it so it gets uploaded to the server.
4. Return to the home page, and now load this new product.
5. Try to rename it, and ensure it works properly.
6. Now, try to load an existing product.
7. Try to rename it and ensure that everything works here as well. Also, check that the previous editing history has been wiped.
8. Try whatever you want to break this otherwise!

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487)

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -45,7 +45,7 @@ app.use(cors());
 // POST requests made to /upload will be handled here.
 app.route(ROUTE_PREFIX + '/upload').post(function (req, res, next) {
     // TODO: Putting this on the header isn't great. The body has the zipped folder. And usernames in the URL doesn't look great either. Maybe improve this somehow.
-    const user = req.headers.user 
+    const user = req.headers.user;
     if (!user) {
         // Send back error if the user uploading the storyline was not provided.
         responseMessages.push({
@@ -72,11 +72,7 @@ app.route(ROUTE_PREFIX + '/upload').post(function (req, res, next) {
     // Upload the file to the server, into the /files/ folder.
     form.parse(req, async function (err, field, file) {
         if (err) {
-            responseMessages.push({
-                type: 'WARNING',
-                message: 'Upload Aborted: an error has occurred while parsing the uploaded form: ' + err
-            });
-            logger('WARNING', 'Upload Aborted: an error has occurred while parsing the uploaded form: ', err);
+            logger('WARNING', 'Upload Aborted: an error has occurred while parsing the uploaded form: ' + err);
             res.status(500).send({ status: 'Internal Server Error' });
             return;
         }
@@ -90,10 +86,6 @@ app.route(ROUTE_PREFIX + '/upload').post(function (req, res, next) {
         //if (!projectNameRegex.test(fileName)) {
         // SECURITY FEATURE (temporary): Make sure the project name isn't `scripts`, or `help`, and doesn't contain . or / in order to prevent overwriting folders.
         if (fileName !== 'scripts' && fileName !== 'help' && !fileName.includes('/') && !fileName.includes('.')) {
-            responseMessages.push({
-                type: 'WARNING',
-                message: 'Upload Aborted: file does not match Storylines UUID standards.'
-            });
             logger('WARNING', 'Upload Aborted: file does not match Storylines UUID standards.');
             // Delete the uploaded zip file.
             safeRM(secureFilename, UPLOAD_PATH);
@@ -103,12 +95,10 @@ app.route(ROUTE_PREFIX + '/upload').post(function (req, res, next) {
 
         // Before unzipping, create the product folder in /public/ if it doesn't exist already.
         if (!fs.existsSync(fileName)) {
-            responseMessages.push({ type: 'INFO', message: `Successfully created new product ${fileName}` });
             logger('INFO', `Successfully created new product ${fileName}`);
             fs.mkdirSync(fileName);
             newStorylines = true;
         }
-
         // Unzip the contents of the uploaded zip file into the target directory. Will overwrite
         // old files in the folder.
         decompress(secureFilename, fileName).then(async () => {
@@ -118,15 +108,13 @@ app.route(ROUTE_PREFIX + '/upload').post(function (req, res, next) {
             // files.forEach((file) => {
             //     validateFile(file, fileName);
             // });
-            responseMessages.push({ type: 'INFO', message: `Uploaded files to product ${fileName}` });
             logger('INFO', `Uploaded files to product ${fileName}`);
             // Initialize a new git repo if this is a new storyline.
             // Otherwise, simply create a new commit with the zipped folder.
             if (!newStorylines) {
-                await commitToRepo(fileName, user, false)
-            }
-            else {
-                await initGitRepo(fileName, user)
+                await commitToRepo(fileName, user, false);
+            } else {
+                await initGitRepo(fileName, user);
             }
             // Finally, delete the uploaded zip file.
             safeRM(secureFilename, UPLOAD_PATH);
@@ -143,10 +131,10 @@ app.route(ROUTE_PREFIX + '/upload').post(function (req, res, next) {
 // GET requests made to /retrieve/ID/commitHash will be handled here.
 // Calling this with commitHash as "latest" simply fetches the product as normal.
 app.route(ROUTE_PREFIX + '/retrieve/:id/:hash').get(function (req, res, next) {
-    // This user is only needed for backwards compatibility. 
+    // This user is only needed for backwards compatibility.
     // If we have an existing storylines product that is not a git repo, we need to initialize a git repo
     // and make an initial commit for it, but we need the user for the commit.
-    const user = req.headers.user 
+    const user = req.headers.user;
     if (!user) {
         // Send back error if the user uploading the storyline was not provided.
         responseMessages.push({
@@ -161,7 +149,7 @@ app.route(ROUTE_PREFIX + '/retrieve/:id/:hash').get(function (req, res, next) {
     var archive = archiver('zip');
     const PRODUCT_PATH = `${TARGET_PATH}/${req.params.id}`;
     const uploadLocation = `${UPLOAD_PATH}/${req.params.id}-outgoing.zip`;
-    const commitHash = req.params.hash
+    const commitHash = req.params.hash;
 
     // Check if the product exists.
     if (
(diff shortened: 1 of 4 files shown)


mohsin-r (changes requested):
Since we are renaming files on UUID change, we also have the map configs which have default naming in the format `uuid-map-#.json`. Perhaps its worth it to change those?

> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OEZfPqy4jSaD2IqKNS_:b-sngwdk)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @RyanCoulsonCA)

___
*[`server/index.js` line 347 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-OEZeGCr14K4uPh2uEVj:-OEZeGCr14K4uPh2uEVk:b2q9qws) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/70c571a03b4275ab877cce1ba096344910d92324/server/index.js#L347)):*
> ```JavaScript
>                                 fs.rmSync(NEW_PATH + `/.git`, { recursive: true, force: true });
>                                 const git = simpleGit(NEW_PATH);
>                                 await git.init();
> ```

Curious about the choice of creating a new git repo on rename. If a user just doesn't like their product name and wants a different name, it forces them to nuke the version history. Feel free to let me know if I'm missing something from a technical standpoint.
___
*[`server/index.js` line 502 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-OEZe3Wp6l7ed_NZ6qrM:-OEZe3Wp6l7ed_NZ6qrN:b3whpmv) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/70c571a03b4275ab877cce1ba096344910d92324/server/index.js#L502)):*
> ```JavaScript
> 
>     // Push to responseMessages.
>     responseMessages.push({ type: type, message: message });
> ```

Smart coding! :brain:

RyanCoulsonCA (commented):
Good catch, working on this now.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OE_LaLT-W2VvFe9CaVY:bef78yh)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @mohsin-r)

___
*[`server/index.js` line 347 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-OEZeGCr14K4uPh2uEVj:-OE_LfMGDbyrPQerQbGo:beji33o) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/70c571a03b4275ab877cce1ba096344910d92324/server/index.js#L347)):*

Curious about the choice of creating a new git repo on rename. If a user just doesn't like their product name and wants a different name, it forces them to nuke the version history. Feel free to let me know if I'm missing something from a technical standpoint.
</blockquote></details>

If we don't wipe the Git history, then we'd basically need to add a bunch of new processes to handle reverting a rename:

If a user renames a product and then reverts it, then everything within the product folder will revert back to using the previous UUID (except the folder name itself), which will mean we'd need to add another function to the server to rename the folder back to the previous name. Additionally, I assume we'd also need to fire another request at the database to revert the UUID there as well.

If we really want to keep the Git history we can definitely implement all of this, but I just made the assumption that we'd want to choose the easier option here. We can also add a warning message on the rename screen to let the user know that the edit history will be wiped.

mohsin-r (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @RyanCoulsonCA)

___
*[`server/index.js` line 347 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-OEZeGCr14K4uPh2uEVj:-OE_dlnECVqeEaTwCFQh:bm6ywr2) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/70c571a03b4275ab877cce1ba096344910d92324/server/index.js#L347)):*

If we don't wipe the Git history, then we'd basically need to add a bunch of new processes to handle reverting a rename:

If a user renames a product and then reverts it, then everything within the product folder will revert back to using the previous UUID (except the folder name itself), which will mean we'd need to add another function to the server to rename the folder back to the previous name. Additionally, I assume we'd also need to fire another request at the database to revert the UUID there as well.

If we really want to keep the Git history we can definitely implement all of this, but I just made the assumption that we'd want to choose the easier option here. We can also add a warning message on the rename screen to let the user know that the edit history will be wiped.
</blockquote></details>

Yeah that totally makes sense to me now, I hadn't considered that before. I think a warning message would be helpful for sure.

RyanCoulsonCA (commented):
Donethanks.

Because the `rename` endpoint only takes the two renamed config files as opposed to the entire ZIP structure, this had to be implemented on the server-end. I'm not a huge fan of this as it introduces more blocking calls, but I think that since renaming shouldn't be used very often it's not a huge deal.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OF8d65T4xR5FMWGx_wX:btpuo8r)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @mohsin-r)

IshavSohal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @mohsin-r and @RyanCoulsonCA)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-OFT6JZe5G0EBRkWJW7j:-OFT6JZe5G0EBRkWJW7k:b-4221pt) (no related file):*
When a product containing a map has its UUID changed and the product is then loaded again, I noticed that the map fails to render. However upon reloading, the map renders just fine. In the console I get the following error when I open the map panel:
```
TypeError: this.rampEditorApi.getConfig is not a function
```

RyanCoulsonCA (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @IshavSohal and @mohsin-r)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-OFT6JZe5G0EBRkWJW7j:-OFgmuuw09TwRbF9shls:bdnmwe8) (no related file):*

When a product containing a map has its UUID changed and the product is then loaded again, I noticed that the map fails to render. However upon reloading, the map renders just fine. In the console I get the following error when I open the map panel:
```
TypeError: this.rampEditorApi.getConfig is not a function
```
</blockquote></details>

Good catch. A side effect of doing the file renaming on the server-end is that the map configuration files in the ZIP file weren't being renamed until the product was re-fetched from the server.

I've added a new function to rename the files in the ZIP file locally as well to fix this. I think it would be ideal in the future to drop the server-side renaming and just keep the new client side changes, but this would once again require us to send the entire modified ZIP file back to the server when renaming which would make it take a lot longer.

IshavSohal (approved):
> Reviewed 1 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

mohsin-r (approved):
> Reviewed 1 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

szczz (changes requested):
I pulled this branch locally and moved the server folder outside of storylines-editor, but it still seems to be hanging on rename. It doesn't let me rename a storyline to something that already exists, but pressing rename for a valid uuid results in an infinite loading wheel. I don't see anything in the console and I don't see any requests being made in the network tab.



Edit: Just to clarify, I also tested this with nothing accessing the folders other than the terminals running the express server and vue app

> Reviewed 1 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OGFy2QWDEeDT4qwMcp0:b-5zo1ni)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

RyanCoulsonCA (commented):
Very interesting. I haven't been able to reproduce the hanging in any case other than a file being in use somehow (even subtly, sometimes even if I close a file in VS Code and then try renaming, it will still block until I close VS Code). 

I've done a bunch of research online to see if there's a way to detect if the `rename` call is being blocked by something, but I can't seem to find anything that will be helpful here.

Either way, I've added in a `catch` block on the front end to detect 500 errors returned from the server (which probably should've been done in the first place). If one does occur, an error will pop up saying that the rename has failed, and none of the renaming steps will take place. This should resolve the issue with the infinite spinner.

Try to test this again with these new changes and if the server hangs, wait a few minutes to see if an `EPERM` error pops up in the console (which is the error that usually pops up for me if the `rename` call is blocked).

For further debugging, I've added a slightly modified `index.js` below with some conveniently placed `console.log`s so you can see where it's hanging. 
[index.zip](https://github.com/user-attachments/files/18381569/index.zip)


> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OGGgXRWEBdwcj1TbtnD:b-r7bqi5)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @IshavSohal, @mohsin-r, and @szczz)

yileifeng (commented):
Working well for all cases tested :+1:

One small issue found is that if you load in a product, modify the UUID input field and proceed to rename, the functionality is missing. Maybe disabling the feature or ignoring the modified input could be a workaround. 

> Reviewed 1 of 2 files at r2, 1 of 3 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OGV-kpZEn_gwaW6e7dd:bgslqc3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

RyanCoulsonCA (commented):
I've modified the code to use the currently loaded UUID (from `this.configFileStructure.uuid`) instead of using the value that was typed into the UUID input which should fix this.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OGVJRk46NA6T11fVZ5z:b-kgrj1l)* status: 3 of 4 files reviewed, all discussions resolved (waiting on @yileifeng)

szczz (approved):
Frontend is still hanging for me and displaying an infinite loading wheel but I can see the rename request in the network tab and the product is successfully renamed... Likely just an issue on my end. 

> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487#-:-OGVvmGi5JivUzUKoLzR:b-mkrfan)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

yileifeng (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/487)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RyanCoulsonCA)

Issue #454, opened 2024-11-28
Make a map, clone a map, delete a map, break config editor


dan-bowerman:
## Description

If i make a map, it creates a RAMP config, for this example we'll call it `bowerlines/ramp-config/bowerlines-map-1.json`. 

## Environment

- All

## Screenshots
![image](https://github.com/user-attachments/assets/5ad35839-9671-4c16-9bfb-545851f3b6b5)

## Steps to reproduce

1. Create a map
2. Use the "copy" tool to duplicate the slide
3. Delete or modify the type of the first map
4. The copied slide is looking for the original map config, which no longer exists
5. The RAMP4 config editor does not load and console errors are thrown

## Expected behavior

- An error, warning, or fallback if the map config isn't present, and/or
- A copied RAMP config when cloning a slide so there isn't reliance on the initial config. No use-case should see two panels relying on the same map config! 

## Actual behavior

I can't access the map config, I can't load the editor, and I have to either clear the problem with the advanced tab or delete the slide and start over.

PR #497, opened 2024-12-20
Ensure 'copy slide' always makes new RAMP config

gordlin:
### Related Item(s)
Issue #454 

### Changes
- [Fix] Ensures a new RAMP config file is created every time a slide with a map panel is copied, so there aren't any shared configs between slides.

### Notes
It doesn't seem we're currently _deleting_ config files when a map panel-containing slide is deleted. Should we do that?

### Testing
Steps:
1. Go to any product (`00000000-0000-0000-0000-000000000000`).
2. Create or go to a slide with a map panel. In the RAMP editor, make a few changes from the default.
3. Go to the advanced editor and note the filename of the config.
4. Copy the slide. 
5. Open the new slide, go to the map panel, the select the advanced editor. The config filename should be different from the original slide.
6. Go back to the copied map panel. In the RAMP editor, check the items you changed in step 2. They should still be changed here, even if you didn't press 'Save changes'.


IshavSohal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @gordlin)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OFiMdIDB5ELNsfIGmiQ:-OFiMdIDB5ELNsfIGmiR:bhvrlto) (no related file):*
The fix works for map panels, but the issue still occurs when the maps are within a slideshow panel.

Not exactly related, but I noticed that an issue similar to #454 occurs for assets and charts. For those, I think you can increment the source count of each asset/chart of the copied slide within `copySlide()`, rather than create new files. 

Regarding deleting the config files, there's a fix for that in #343

RyanCoulsonCA (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @gordlin and @IshavSohal)

___
*[`src/components/slide-toc.vue` line 661 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OG_oX9O0WiaM09iAZzj:-OG_oX9PAw42MrzITh-g:bz327je) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/6724aa4d65732bd19ae445201dc15aad8a610ee3/src/components/slide-toc.vue#L661)):*
> ```Vue
>         // Nested setTimeout calls may not be the best option. It works though.
>         // Maybe try replacing with promise chain?
>         setTimeout(async () => {
> ```

I'd say we should go with Promises instead of the nested timeouts. While it's likely everything will be done before the timeout triggers, we can't guarantee it and it can lead to some hard-to-find bugs later on.

gordlin (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @IshavSohal and @RyanCoulsonCA)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OFiMdIDB5ELNsfIGmiQ:-OU2H79U7UBkVShB0wXX:brbwecq) (no related file):*
> the issue still occurs when the maps are within a slideshow panel.

Donethanks!
___
*[`src/components/slide-toc.vue` line 661 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OG_oX9O0WiaM09iAZzj:-OU2HTmz7nYf8GIurN5Q:b1b0tjs) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/6724aa4d65732bd19ae445201dc15aad8a610ee3/src/components/slide-toc.vue#L661)):*

I'd say we should go with Promises instead of the nested timeouts. While it's likely everything will be done before the timeout triggers, we can't guarantee it and it can lead to some hard-to-find bugs later on.
</blockquote></details>

Nested timeouts are gone now, but for some reason the `copySlide` function still needs a setTimeout for switching slides, else it'll fail to switch. Quite a persistent bug; if it's fine with you, I'll open a new issue for it.
___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OU2H0VWDBWCvExPVHWT:-OU2H0VWDBWCvExPVHWU:b541qo8) (no related file):*
Completely rewrote the code for this PR. The solution is a lot more robust, and the various edge cases are much better handled.

IshavSohal (approved):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: all files reviewed, 2 unresolved discussions (waiting on @RyanCoulsonCA)

szczz (commented):
Failed to save changes after following the testing steps on the ```00000000-0000-0000-0000-000000000000``` product. Saving without changes or minor text changes works fine.

 Console grouse below. 



> Reviewed 1 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-:-OWucmmp311Kp7_2us0E:b-fsoa8f)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @IshavSohal and @RyanCoulsonCA)

spencerwahl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @IshavSohal, @RyanCoulsonCA, and @szczz)

___
*[`src/components/slide-toc/slide-toc.vue` line 669 at r5](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OXi-6ud8UoRd4vE8-H4:-OXi-6ud8UoRd4vE8-H5:b-9zmr4r) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/a1a807c04ecd496db69d85e1f8d676a11423c1af/src/components/slide-toc/slide-toc.vue#L669)):*
> ```Vue
>         Message.success(this.$t('editor.slide.copy.success'));
> 
>         // TODO: setTimeout appears to be necessary for now, otherwise the slide switch fails to occur. Investigate & replace.
> ```

Maybe `await nextTick()` would work?

gordlin (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @IshavSohal, @RyanCoulsonCA, @spencerwahl, and @szczz)

___
*[`src/components/slide-toc/slide-toc.vue` line 669 at r5](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OXi-6ud8UoRd4vE8-H4:-OY6hZoI7hX7Ip4OL3Rn:b-j4w3ko) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/a1a807c04ecd496db69d85e1f8d676a11423c1af/src/components/slide-toc/slide-toc.vue#L669)):*

Maybe `await nextTick()` would work?
</blockquote></details>

Didn't work, unfortunately. I did find another solution, though: awaiting a `Promise` on the emitted event where the resolve is "sent up the chain" to the parent, where the final parent of the chain will `resolve` the whole chain. 

This appears to work well, and this approach could likely be used to replace all `setTimeout`s in the editor. However, with all cross-component variables soon being put into Pinia stores in other PRs, I think this may not be necessary; doing that gets rid of many emits, possibly making most of the code synchronous already and eliminating the need for this. Worth looking into later (#515) once all those PRs are in. 

Donethanks!

gordlin (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @IshavSohal, @RyanCoulsonCA, @spencerwahl, @szczz, and @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497#-OYnSpsW8RHpFDENzBIK:-OYnSpsW8RHpFDENzBIL:b-ywu5z8) (no related file):*
Rebased with main, fixed merge conflicts. Behaviour should still be the same. All ready!

szczz (approved):
> @szczz reviewed 1 of 4 files at r2, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/497)* status: all files reviewed, 3 unresolved discussions (waiting on @IshavSohal, @RyanCoulsonCA, @spencerwahl, and @yileifeng)

Issue #520, opened 2025-01-23
'Session Extended' toast appears on mouse movement in preview


IshavSohal:
## Description

Upon previewing a product, I noticed that the 'Session Extended' toast will appear each time mouse movement is detected. However this only occurs upon switching languages and then switching back

## Environment

Reproduced in Chrome, Firefox and Edge

## Steps to reproduce

1. Load in any product
2. Open the preview
3. Switch languages twice
4. Notice that the 'Session Extended' toast appears whenever you move the mouse, which clogs up the screen

## Expected behavior

Toast should not appear on mouse movement

## Actual behavior

Toast appears on mouse movement

## Possible solution

Upon switching languages twice, perhaps some default value gets set, and causes mouse movement to trigger the toast. Not exactly sure though.

PR #531, opened 2025-01-30
Close BroadcastChannel upon lang change in preview

IshavSohal:
### Related Item(s)
#520

### Changes
- ~~Persist the state of the lock store between tabs, in particular the editor and preview tabs~~
- Close the preview's BroadcastChannel upon lang change (this was causing the reported error)
- Include `secret` within the window props upon lang change in the preview

### Notes
<del> -  I decided to use [this](https://github.com/prazdevs/pinia-plugin-persistedstate) plugin to persist the state of the lock store. If we don't want to use this, let me know of preferred alternatives. I could also try creating our own composable if needed</del>
<del>- I noticed that, within the preview component, when `window.props` is not defined (which occurs upon reloading the tab), the BroadcastChannel was not being created. Was this intentional? Currently I'm creating the BroadcastChannel in both cases (hence the plugin), but I can revert this if necessary </del>

### Testing
Steps:
1. Load any product
2. Open devtools
3. Within the metadata editor, preview the product
4. Within the preview tab open devtools
5. Move your mouse around in the preview tab
6. Switch back to the editor tab, and notice the console logs indicating that the messages sent by the preview's BroadcastChannel were received
7. Within the preview tab switch languages and repeat 5-6
8. Within the preview tab refresh and repeat 5-6
9. Throughout the testing, the 'Session Extended' toast should not appear on mouse movement in the preview tab

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531)

### DIFF
--- a/src/components/editor.vue
+++ b/src/components/editor.vue
@@ -774,6 +774,7 @@ export default class EditorV extends Vue {
             return slide ?? JSON.parse(JSON.stringify(this.defaultBlankSlide));
         });
         const lockStore = useLockStore();
+
         setTimeout(() => {
             const routeData = this.$router.resolve({
                 name: 'preview',
--- a/src/components/preview.vue
+++ b/src/components/preview.vue
@@ -113,6 +113,7 @@ export default class StoryPreviewV extends Vue {
     totalTime = import.meta.env.VITE_APP_CURR_ENV ? Number(import.meta.env.VITE_SESSION_END) : 30;
 
     extendSession(showPopup?: boolean): void {
+        // Only send message to other BroadcastChannel if preview is connected to editor
         if (!this.savedProduct) {
             this.broadcast?.postMessage({ action: 'extend', showPopup });
         }
@@ -128,7 +129,7 @@ export default class StoryPreviewV extends Vue {
             this.config = JSON.parse(JSON.stringify(window.props.configs[this.lang]));
             this.configs = window.props.configs;
             this.configFileStructure = window.props.configFileStructure;
-            // this broadcast channel will be used to communicate regarding sessions with the main editor tab
+            // This broadcast channel will be used to communicate regarding sessions with the main editor tab
             this.broadcast = new BroadcastChannel(window.props.secret);
             this.broadcast.onmessage = (e) => {
                 const msg = e.data;
@@ -252,7 +253,7 @@ export default class StoryPreviewV extends Vue {
                 fetch(this.apiUrl + `/retrieveMessages`).then((res: any) => {
                     axios
                         .post(import.meta.env.VITE_APP_NET_API_URL + '/api/log/create', {
-                            messages: res.data.messages
+                            messages: res.data?.messages
                         })
                         .catch((error: AxiosError) => console.log(error.response || error));
                 });
@@ -278,17 +279,22 @@ export default class StoryPreviewV extends Vue {
 
     // reload preview page with FR config
     changeLang(): void {
+        this.broadcast?.close();
         const newLang = this.lang === 'en' ? 'fr' : 'en';
         const routeData = this.$router.resolve({
             name: 'preview',
             params: { lang: newLang, uid: this.uid }
         });
+        const secret = window.props.secret;
+        const timeRemaining = window.props.timeRemaining;
 
         // update window props on refresh (to prevent having to fetch from server again)
         const refreshTab = window.open(routeData.href, '_self');
         (refreshTab as Window).props = {
             configs: this.configs,
-            configFileStructure: this.configFileStructure
+            configFileStructure: this.configFileStructure,
+            secret,
+            timeRemaining
         };
         this.$forceUpdate();
     }
--- a/src/stores/lockStore.ts
+++ b/src/stores/lockStore.ts
@@ -78,7 +78,6 @@ export const useLockStore = defineStore('lock', {
                         this.uuid = uuid;
                         this.secret = data.secret;
                         this.broadcast = new BroadcastChannel(data.secret);
-
                         resolve();
                     }
                 };


mohsin-r (commented):
I think we don't need the `BrodcastChannel` in the refresh case because we treat that as a "user loaded preview independently" of the main tab and hence it should not track activity in the main editor tab. 

However, this does reopen the case of the language switch and what to do there.

One option to really simplify things is to remove the activity tracker in the preview tab. While this makes it somewhat inconvenient if a user times out while they're looking at their glorious preview, because we now autosave on timeout, they'd just have to lock the storyline again and they'd be able to continue.

Another option is to investigate what causes the toast to trigger when switching languages and stop it there.

Yet a third option is to overrule the idea of a refresh being a new load, and also track activity there, in which case your code is perfect.

Will tag on the meeting topic to this one and see what people think.

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531#-:-OJ09rod9vzj0YNXYvCN:bgdx9m6)* status: all files reviewed, 2 unresolved discussions (waiting on @IshavSohal)

___
*[`src/components/metadata-editor.vue` line 840 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531#-OJ04wb0BSW9_Pm2wu7F:-OJ04wb0BSW9_Pm2wu7G:bm17kd6) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/3311873e3376a0536fe9c07379c54be252c4a98d/src/components/metadata-editor.vue#L840)):*
> ```Vue
>                     const msg = e.data;
>                     if (msg.action === 'extend') {
>                         console.log('metadata - received extend message from preview');
> ```

Blocker for console logs before merging
___
*[`src/components/preview.vue` line 207 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531#-OJ0957F2SbCEdOW4JSr:-OJ0957F2SbCEdOW4JSs:b-e91c9c) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/3311873e3376a0536fe9c07379c54be252c4a98d/src/components/preview.vue#L207)):*
> ```Vue
>         }
> 
>         // This broadcast channel will be used to communicate regarding sessions with the main editor tab
> ```

So if we put this code block here, it'll also trigger a broadcast channel in the case that someone loads the preview page independently (i.e. without the main editor tab) and in that case there are no sessions to extend or handle.

IshavSohal (commented):
> I think we don't need the `BroadcastChannel` in the refresh case because we treat that as a "user loaded preview independently" of the main tab and hence it should not track activity in the main editor tab.

Ah I see now, considered a refresh as an independent load makes sense. If we want to keep this the way it is, I can move the BroadcastChannel back  to the `window.props` case. Also, we won't need the plugin anymore, since we can just pass the secret to the window in `changeLang()`.

> Another option is to investigate what causes the toast to trigger when switching languages and stop it there.

The reason why the `Session Extended` toast appears on mouse move is because, upon switching languages, `window.props` doesn't contain the `secret` (see the `changeLang()` method). So, the name provided for the BroadcastChannel defined in the `preview` (`undefined`) would be different from the name of the one defined in the `lockStore`, and thus would not be connected to one another. 

In addition, BroadcastChannels defined in the `preview` are not being closed upon a lang change. So, once two lang changes occur, there will be two BroadcastChannels with a name of `undefined`, and these begin to communicate with each other.

We can fix the toast issue by closing the BroadcastChannel in `changeLang()`

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531#-:-OJ5Gmm6ALhE75lMRauX:bilhadv)* status: all files reviewed, 2 unresolved discussions (waiting on @IshavSohal)

mohsin-r (commented):
Ah I see, either removing the toast or making the channel persist on language change seem like good suggestions. I'm in favour of the latter as a user I would still want my session to persist if I'm just previewing the other language.

Having said all of that, I still think its worth holding off on implementing anything now and instead discussing below/in the next meeting whether to remove the broadcasts/activity tracking in the preview tab as it simplifies the code greatly and only takes away what is a very niche benefit. If people are in favour of keeping the activity tracking, then we should probably keep hagglin' on what makes sense to do in a language change.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531#-:-OJKPgR87-eyQH9DA2sn:b-3j492z)* status: all files reviewed, 2 unresolved discussions (waiting on @IshavSohal)

IshavSohal:
Now the `BroadcastChannel` is only defined when `window.props` is defined. The toast issue has been fixed by closing the existing `BroadcastChannel` upon a lang change, as well as including the `secret` within the window props. I've also removed the plugin, since we are considering a preview tab to be independent from the editor upon a refresh.

mohsin-r (commented):
Sorry for blocking this for so long. Seems to be working well now. Please do a rebase whenever you can as there seem to be a bunch of unrelated "changes" showing up which would undo later work.

> Reviewed 3 of 6 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531#-:-OLuSWWG0hAiO8kimZJp:b-lvdx0a)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @IshavSohal)

mohsin-r (approved):
Heh reviewable was showing weird stuff, all good! Just needs console log removal.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531#-:-OM7GThG11qhLKdWb5Kj:bt4whqs)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @IshavSohal)

mohsin-r (approved):
> Reviewed 1 of 6 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

spencerwahl (approved):
> Reviewed 3 of 6 files at r2, 1 of 3 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

szczz (approved):
> Reviewed 3 of 6 files at r2, 1 of 3 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/531)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IshavSohal)

Issue #491, opened 2024-12-18
Remove redundant schema file and ensure editor specific flags are removed on save


szczz:
## Description

Discussed in the 12-17 Platform meeting

We have begun adding editor specific flags to the schema which is leading to a mismatch with the storylines project. The intention was that we would use the schema from storylines and sync our copy whenever changes were made. 

@yileifeng has created a [discussion ](https://github.com/ramp4-pcar4/storylines-editor/discussions/490) that further describes the issue. 

We need to update RESPECT with the following changes 

1. Remove StorylinesSchema.json file in the editor codebase (verify it is unused)
2. Ensure any editor specific flags (modified and other temporary panel editor properties) are not included in the config when saved (check the saveChanges function of each panel component)

## Related issues

#490

PR #532, opened 2025-02-10
Remove redundant schema file, Remove editor-exclusive props from schema

IshavSohal:
### Related Item(s)
#491

### Changes
- Removed the redundant `StorylinesSchema.json` file, since it was not being used anywhere else
- Prevented the `slide-editor` component from adding the following props to a panel
  -  `modified`
  - `rightOnly`
  - `centerPanel`
  - `centerSlide`
- Prevented the `image-editor` component from adding the `id` prop to ImagePanel's within a slideshow
- Removed the above props and  `userCreated` from the schema
- Called `saveChanges()` before executing main logic of `panelModified()` method, so that unsaved changes are captured and used when comparing the provided panel with its corresponding default config
- Fixed a few TS errors
- Fixed map src values for product in `public` folder (was getting errors when testing)
- Added `cssClasses` to the editor's schema
- Upon loading a product in (either new or existing), a `styles/main.css` file will be added to `configFileStructure` (and later the product itself upon saving), which contains classes used for styling panels upon toggling the `center slide` and `center panel` controls
- Added two new interfaces: `SlideshowImagePanel` and `SlideshowChartPanel`. Also added two new panel types: `slideshowImage` and `slideshowChart` 
  - These are the same as regular slideshow panels, except that their `items` are restricted to `ImagePanel`'s and `ChartPanel`'s
  - Also updated the schema for `multimediaSlideshow`

### Note
- I noticed that, for map panels, then `timeSlider` prop exists in the editors schema but not in the storylines schema. However it seems like the `timeSlider` prop is used within storylines. Should this be added to the storylines schema, or removed from the editor schema?

### Testing
Steps:
1. Create a new product
2. Create a new slide
3. Change the type of a panel to a media panel (ex. Image, Video, Chart)
4. Notice there is no warning modal
5. Add some content to the panel (without saving)
6. Switch the type to another type, and observe the warning modal
7. Toggle the 'fullscreen panel' control, and go to the advanced editor. Observe that the 'rightOnly' prop is not added to the panel
8. Toggle the 'center panel content' control, and go to the advanced editor. Observe that the 'centerPanel' prop is not added to the panel
9. Toggle the 'center slide content' control, and go to the advanced editor. Observe that the 'centerSlide' prop is not added to the panel
10. Create an image panel, and upload more than one image
11. Go to the advanced editor, and observe that the `id` prop was not added to any of the image panels

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)


yileifeng (changes requested):
> Reviewed 3 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: 3 of 9 files reviewed, 2 unresolved discussions

___
*[`public/StorylinesSlideSchema.json` line 414 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OIprITqAsVhOhlsRnjA:-OIprITqAsVhOhlsRnjB:bmm7urr) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/d50eaad674628eaba4ec57bf1260a45f0557f427/public/StorylinesSlideSchema.json#L414)):*
> ```JSON
>                     "description": "A caption to display below the slideshow."
>                 },
>                 "userCreated": {
> ```

This looks like another temporary schema property exclusive to RESPECT that should be discarded upon saving 
___
*[`public/StorylinesSlideSchema.json` line 516 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OIpspug4QWdzRF8FtK1:-OIpspug4QWdzRF8FtK2:b-15zb88) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/d50eaad674628eaba4ec57bf1260a45f0557f427/public/StorylinesSlideSchema.json#L516)):*
> ```JSON
>             "default": true
>         },
>         "rightOnly": {
> ```

Same with "rightOnly", "centerSlide", "centerPanel" which are only used in slide editor formatting

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: 1 of 13 files reviewed, 3 unresolved discussions (waiting on @yileifeng)

___
*[`public/StorylinesSlideSchema.json` line 414 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OIprITqAsVhOhlsRnjA:-OJ5DRh10dAyR2iXhIVu:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/d50eaad674628eaba4ec57bf1260a45f0557f427/public/StorylinesSlideSchema.json#L414)):*

This looks like another temporary schema property exclusive to RESPECT that should be discarded upon saving
</blockquote></details>

Donethanks
___
*[`public/StorylinesSlideSchema.json` line 516 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OIpspug4QWdzRF8FtK1:-OJ5DT9y2D_4ZW80iMVD:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/d50eaad674628eaba4ec57bf1260a45f0557f427/public/StorylinesSlideSchema.json#L516)):*

Same with "rightOnly", "centerSlide", "centerPanel" which are only used in slide editor formatting
</blockquote></details>

Donethanks
___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OJ5DV3e9Vtz2DhntRAJ:-OJ5DV3e9Vtz2DhntRAK:b8ikx7r) (no related file):*
Now that the `centerPanel` and `centerSlide` properties are not added to panels, the  `centerPanel` and `centerSlide` props, within the `slide-editor`, are set by checking the `customStyles` prop of the panel
  - If both panels have `text-align: center;` set, then `centerPanel` is set to true, otherwise false
  - If left panel has `text-align: right;` set and right panel has `text-align: left` set , then `centerSlide` is set to true, otherwise false

However, when there is a fullscreen panel, both the `center slide content` control and the `center panel control` perform the exact same action, which is to set  `text-align: center;`  on the individual panel
  - Within `onSlideChanges()` there is no way to distinguish between whether the fullscreen panel has had the `center slide content` control enabled or the `center panel content` control enabled
  - So, I decided to disable the `center slide content` control when there is a fullscreen panel, and to set  `centerPanel`  to true when the individual panel has `text-align: center;` set
 
Let me know your thoughts on this design decision, and if you have any suggestions. 

The main issue I have with this is that if the user enables the `center slide content` control, and then enables the `fullscreen panel` control, and then disables the `fullscreen panel` control, the `center panel content` control would be enabled, as opposed to the `center slide content` control. If the user wants the `center slide content` control to be enabled the entire time, they would have to manually set it again.

yileifeng (changes requested):
> I noticed that, for map panels, then `timeSlider` prop exists in the editors schema but not in the storylines schema. However it seems like the `timeSlider` prop is used within storylines. Should this be added to the storylines schema, or removed from the editor schema?

Seems like it should just be a property added to the storyline schema 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-:-OJJYElY3CpPNKyXO-vk:b-7vy9yt)* status: 1 of 13 files reviewed, 4 unresolved discussions (waiting on @IshavSohal)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OJ5DV3e9Vtz2DhntRAJ:-OJJYLlM6XiHhmdj7FiU:bo2m8im) (no related file):*

Now that the `centerPanel` and `centerSlide` properties are not added to panels, the  `centerPanel` and `centerSlide` props, within the `slide-editor`, are set by checking the `customStyles` prop of the panel
  - If both panels have `text-align: center;` set, then `centerPanel` is set to true, otherwise false
  - If left panel has `text-align: right;` set and right panel has `text-align: left` set , then `centerSlide` is set to true, otherwise false

However, when there is a fullscreen panel, both the `center slide content` control and the `center panel control` perform the exact same action, which is to set  `text-align: center;`  on the individual panel
  - Within `onSlideChanges()` there is no way to distinguish between whether the fullscreen panel has had the `center slide content` control enabled or the `center panel content` control enabled
  - So, I decided to disable the `center slide content` control when there is a fullscreen panel, and to set  `centerPanel`  to true when the individual panel has `text-align: center;` set
 
Let me know your thoughts on this design decision, and if you have any suggestions. 

The main issue I have with this is that if the user enables the `center slide content` control, and then enables the `fullscreen panel` control, and then disables the `fullscreen panel` control, the `center panel content` control would be enabled, as opposed to the `center slide content` control. If the user wants the `center slide content` control to be enabled the entire time, they would have to manually set it again.
</blockquote></details>

Initially I was just thinking we would only require removing these exclusive properties upon save so the editor code functionality would remain untouched, but since you found an alternative method to replace these config properties altogether that's even better. 

As for the approach, I think it makes sense to disable the `center slide content` control on a fullscreen panel slide. For the edge case where the user goes from 2-panel --> fullscreen panel --> back to 2-panel setup, would it make sense to add a CSS class `'center-slide'` and `'center-panel'` to track which styles should be applied. This way, even when we return to the 2-panel slide we can check for this class to determine which styles should be applied. Maybe this does not make sense tho as I am not the most familiar with how the CSS works for this part of the code. 
___
*[`src/components/slide-editor.vue` line 677 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OJJjS2j5xC64SlduBpt:-OJJjS2j5xC64SlduBpu:b-hkk7wa) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/392a031a1de661b42477d585f60d18f9b94d109c/src/components/slide-editor.vue#L677)):*
> ```Vue
>     determineEditorType(panel: BasePanel): string {
>         if (panel.type !== PanelType.Slideshow) return panel.type;
>         if ((panel as SlideshowPanel).items.length === 0 || (panel as SlideshowPanel).userCreated)
> ```

Removing the `userCreated` straight up will cause a new bug: when we create a new panel of the slideshow type and add an image, the panel will be detected as an image panel afterward

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: 1 of 13 files reviewed, 4 unresolved discussions (waiting on @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OJ5DV3e9Vtz2DhntRAJ:-OJPny0xCN6q2pDeN__q:b-fzbl7x) (no related file):*

Initially I was just thinking we would only require removing these exclusive properties upon save so the editor code functionality would remain untouched, but since you found an alternative method to replace these config properties altogether that's even better. 

As for the approach, I think it makes sense to disable the `center slide content` control on a fullscreen panel slide. For the edge case where the user goes from 2-panel --> fullscreen panel --> back to 2-panel setup, would it make sense to add a CSS class `'center-slide'` and `'center-panel'` to track which styles should be applied. This way, even when we return to the 2-panel slide we can check for this class to determine which styles should be applied. Maybe this does not make sense tho as I am not the most familiar with how the CSS works for this part of the code.
</blockquote></details>

I noticed that https://github.com/ramp4-pcar4/story-ramp/pull/489 added a `cssClasses` prop to the storylines schema. With this, we can just create a few classes in the storylines `main.css` file, and then assign `cssClasses` to the appropriate class name in the editor. I can try adding this to the editor's schema and using it in `slide-editor`. However, the npm registry for storylines would need to include the classes first.
___
*[`src/components/slide-editor.vue` line 677 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OJJjS2j5xC64SlduBpt:-OJPyexT0OCdPacke00g:b1iawn) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/392a031a1de661b42477d585f60d18f9b94d109c/src/components/slide-editor.vue#L677)):*

Removing the `userCreated` straight up will cause a new bug: when we create a new panel of the slideshow type and add an image, the panel will be detected as an image panel afterward
</blockquote></details>

I see, for this I'll just remove the prop before saving.

IshavSohal:

I noticed that https://github.com/ramp4-pcar4/story-ramp/pull/489 added a `cssClasses` prop to the storylines schema. With this, we can just create a few classes in the storylines `main.css` file, and then assign `cssClasses` to the appropriate class name in the editor. I can try adding this to the editor's schema and using it in `slide-editor`. However, the npm registry for storylines would need to include the classes first.
</blockquote></details>

My mistake, the `main.css` file would actually be within a product, under the `styles` folder. Using the `cssClasses` property would be assuming that the current product has a `styles/main.css` fille, and that file contains the specific classes we want to use (ex. `center-slide`, `center-panel`). I'm not too sure if this assumption is safe to make. 

If we wanted to, we could, upon creating a new storylines, create a `styles/main.css` file, and add the appropriate classes. However this would not fix the issue for existing storylines.

It may just be better to go with your initial solution (removing the properties upon saving), rather than changing the editor functionality.

yileifeng (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: 1 of 13 files reviewed, 4 unresolved discussions (waiting on @IshavSohal)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OJ5DV3e9Vtz2DhntRAJ:-OJshdI0-1hxl-jo6r3F:bnydl4u) (no related file):*

My mistake, the `main.css` file would actually be within a product, under the `styles` folder. Using the `cssClasses` property would be assuming that the current product has a `styles/main.css` fille, and that file contains the specific classes we want to use (ex. `center-slide`, `center-panel`). I'm not too sure if this assumption is safe to make. 

If we wanted to, we could, upon creating a new storylines, create a `styles/main.css` file, and add the appropriate classes. However this would not fix the issue for existing storylines.

It may just be better to go with your initial solution (removing the properties upon saving), rather than changing the editor functionality.
</blockquote></details>

If this becomes too complicated I think a generalized function to remove all temporary config properties would be the easiest path forward

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: 1 of 13 files reviewed, 4 unresolved discussions (waiting on @yileifeng)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OJ5DV3e9Vtz2DhntRAJ:-OK2IUqY33cfV2kzP34_:b-h8liij) (no related file):*

If this becomes too complicated I think a generalized function to remove all temporary config properties would be the easiest path forward
</blockquote></details>

Just realized that we cannot delete the `userCreated` prop from a slideshow panel upon saving, since we need some info to persist within the slideshow panel in order to determine whether it was created explicitly or implicitly. 

An idea I had was to create two new types of panels: `SlideshowImage` and `SlideshowChart`. These would just be slideshow panels whose items are all of one type (all `ImagePanel`'s or all `ChartPanel`'s). We could use these types when creating implicit slideshows, and use the original `Slideshow` type when created explicit slideshows. If this sounds like a good idea I can give this a shot.

IshavSohal:
I've added `cssClasses` to the editor's schema. Now, when a product is loaded in, a `styles/main.css` file will be added to it, which contains the classes corresponding to each of the styles required by the `center slide` and `center panel` controls. This makes setting the controls upon a slide change much easier.

IshavSohal:

Just realized that we cannot delete the `userCreated` prop from a slideshow panel upon saving, since we need some info to persist within the slideshow panel in order to determine whether it was created explicitly or implicitly. 

An idea I had was to create two new types of panels: `SlideshowImage` and `SlideshowChart`. These would just be slideshow panels whose items are all of one type (all `ImagePanel`'s or all `ChartPanel`'s). We could use these types when creating implicit slideshows, and use the original `Slideshow` type when created explicit slideshows. If this sounds like a good idea I can give this a shot.
</blockquote></details>

I've now added in these new types, as well as their corresponding interfaces. I had to update the schema for this to work, meaning that the storylines schema also needs to be updated. I've created a PR for this: https://github.com/ramp4-pcar4/story-ramp/pull/546

Also, styles added to panels via the `cssClasses` property should now be working in the preview. I had to link the stylesheet to the document in the preview tab.

yileifeng (commented):
> Reviewed 2 of 10 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: 3 of 16 files reviewed, 3 unresolved discussions (waiting on @IshavSohal)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532#-OK80EjA9z1hGYIrnSHm:-OL-Pynp6MAjQzhZeVAI:b-ck7zyb) (no related file):*

I've added `cssClasses` to the editor's schema. Now, when a product is loaded in, a `styles/main.css` file will be added to it, which contains the classes corresponding to each of the styles required by the `center slide` and `center panel` controls. This makes setting the controls upon a slide change much easier.
</blockquote></details>

It looks like even with the Storylines package updated, the center panel/slide controls are not applying the styles in the preview

IshavSohal:

It looks like even with the Storylines package updated, the center panel/slide controls are not applying the styles in the preview
</blockquote></details>

Just updated the Storylines version again, the styling should now be working in the preview.

mohsin-r (approved):
> Reviewed 9 of 9 files at r1, 1 of 1 files at r2, 1 of 7 files at r3, 3 of 10 files at r5, 4 of 10 files at r6, 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

szczz (approved):
> Reviewed 1 of 9 files at r1, 1 of 7 files at r3, 3 of 10 files at r5, 4 of 10 files at r6, 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

gordlin (approved):
> Reviewed 1 of 9 files at r1, 1 of 7 files at r3, 3 of 10 files at r5, 4 of 10 files at r6, 7 of 7 files at r7, 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/532)* status: all files reviewed, 1 unresolved discussion (waiting on @yileifeng)

PR #536, opened 2025-02-14
Resolve socket server overload

szczz:
### Changes
- Updates the socket server and client side code to make use of broadcasts. Should resolve the overloading issue.

These changes have been deployed to dev so please test there. 

### Testing
Steps:
1. Ensure you can perform all create, save, rename etc actions on a Storyline
2. A Storyline should not be accessible from more than one tab
3. Please ensure timeout logic works as expected. Currently set to 5 minutes with a 2 min warning.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536)

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -580,24 +580,40 @@ function logger(type, message) {
     console.log(`${currentDate} [${type}] ${message}`);
 }
 
+const clients = new Set();
+
+// Used to broadcast messages to all connected clients
+function broadcastToClients(message){
+    const payload = JSON.stringify(message);
+    clients.forEach((client) => {
+        if(client.readyState === WebSocket.OPEN){
+            logger('INFO', `Payload sent to the client`);
+            client.send(payload);
+        }
+    });
+}
+
 wss.on('connection', (ws) => {
     logger('INFO', `A client connected to the web socket server.`);
+    clients.add(ws);
 
     // The following messages can be received in stringified JSON format:
     // { uuid: <uuid>, lock: true }
     // { uuid: <uuid>, lock: false }
     ws.on('message', function (msg) {
         const message = JSON.parse(msg);
-        const uuid = message.uuid;
+        const {uuid, lock} = message;
+
         if (!uuid) {
             ws.send(JSON.stringify({ status: 'fail', message: 'UUID not provided.' }));
         }
+        logger('INFO', `${msg}`);
         // User wants to lock storyline since they are about to load/edit it.
-        if (message.lock) {
-            // Unlock any storyline that the user was previously locking.
-            delete lockedUuids[ws.uuid];
+        if (lock) {
+            const currentLock = lockedUuids[uuid];
+
             // Someone else is currently accessing this storyline, do not allow the user to lock!
-            if (!!lockedUuids[uuid] && ws.uuid !== uuid) {
+            if (currentLock && ws.uuid !== uuid) {
                 logger('INFO', `A client failed to lock the storyline ${uuid}.`);
                 ws.send(JSON.stringify({ status: 'fail', message: 'Another user has locked this storyline.' }));
             }
@@ -610,6 +626,11 @@ wss.on('connection', (ws) => {
                 lockedUuids[uuid] = secret;
                 ws.uuid = uuid;
                 ws.send(JSON.stringify({ status: 'success', secret }));
+
+                broadcastToClients({
+                    type:'lock',
+                    uuid,
+                });
             }
         } else {
             // Attempting to unlock a different storyline, other than the one this connection has locked, so do not allow.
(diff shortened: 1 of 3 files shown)


mohsin-r (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536)* status: all files reviewed, 3 unresolved discussions (waiting on @szczz)

___
*[`server/index.js` line 586 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536#-OJ4pJLJ6C6FQrCxE1kI:-OJ4pJLKC7s4lrr92QZ2:bqq62a6) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/692ea383eede874c55954f38828206d3fb64ba0a/server/index.js#L586)):*
> ```JavaScript
> 
> // Used to broadcast messages to all connected clients
> function broadcastToClients(message){
> ```

Curious why we need to broadcast a storyline's lock/unlock to all the clients, instead of only the client that requested it (which we already do below with the `ws.send`).
___
*[`server/index.js` line 626 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536#-OJ4ppkDEkAtBQeEA6GI:-OJ4ppkDEkAtBQeEA6GJ:bug183f) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/692ea383eede874c55954f38828206d3fb64ba0a/server/index.js#L626)):*
> ```JavaScript
>                 logger('INFO', `A client successfully locked the storyline ${uuid}.`);
>                 const secret = generateKey();
>                 lockedUuids[uuid] = { secret };
> ```

Since we just have one property now, we could probably go back storing the secret itself instead of an object with the secret inside it. But totally upto you since its a pedantic grouse.
___
*[`src/stores/lockStore.ts` line 82 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536#-OJ4qnqJ6QS1FZL_B4YD:-OJ4qnqJ6QS1FZL_B4YE:b-unrmxx) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/692ea383eede874c55954f38828206d3fb64ba0a/src/stores/lockStore.ts#L82)):*
> ```XML
>             clearInterval(this.timeInterval);
>             if (this.connected) {
>                 this.socket!.send(JSON.stringify({ uuid: this.uuid, user: this.userName, lock: false }));
> ```

I believe the `username` reference should be removed here? Doesn't seem to be anywhere else.

szczz (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @mohsin-r)

___
*[`server/index.js` line 586 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536#-OJ4pJLJ6C6FQrCxE1kI:-OJJHfGL4ilzbMoo9BeE:b-94or8j) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/692ea383eede874c55954f38828206d3fb64ba0a/server/index.js#L586)):*

Curious why we need to broadcast a storyline's lock/unlock to all the clients, instead of only the client that requested it (which we already do below with the `ws.send`).
</blockquote></details>

This is to prevent polling and repeated requests, in a way its a touch of future proofing as well. 

The idea is if all clients are aware that a storyline has been locked, the client won't repeatedly try to access the storyline like we were seeing in the DB (refer to the screenshot I sent on teams). Essentially each client was doing this whenever it requested a locked storyline




Similarly, if we broadcast an unlock message, all clients are aware that a storyline can be accessed.  
___
*[`server/index.js` line 626 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536#-OJ4ppkDEkAtBQeEA6GI:-OJJHdwl2lwqgS5aiUOJ:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/692ea383eede874c55954f38828206d3fb64ba0a/server/index.js#L626)):*

Since we just have one property now, we could probably go back storing the secret itself instead of an object with the secret inside it. But totally upto you since its a pedantic grouse.
</blockquote></details>

Donethanks
___
*[`src/stores/lockStore.ts` line 82 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536#-OJ4qnqJ6QS1FZL_B4YD:-OJJHc5T-dGVzwd18kLi:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/692ea383eede874c55954f38828206d3fb64ba0a/src/stores/lockStore.ts#L82)):*

I believe the `username` reference should be removed here? Doesn't seem to be anywhere else.
</blockquote></details>

Donethanks

yileifeng (approved):
Functionality and code looks fine to me, just logged 2 issues found with renaming that can be addressed in future: #538, #539

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/536#-:-OJJWnZpBRJohuf2GfCp:b-7gjrre)* status: all files reviewed, 3 unresolved discussions (waiting on @mohsin-r)

PR #557, opened 2025-02-21
Fix locking race condition

mohsin-r:
### Changes
- Fixed an issue where one storyline could acquire the secret key of a different storyline due to a race condition, which would prevent it from saving itself.

### Notes
Unfortunately this was being caused by broadcasting the lock/unlock of a storyline to all clients. I believe we had this as a benefit that all clients would be able to know when a storyline was locked/unlocked. However, in certain situations, a client would take the secret key of another client instead of the one they requested. Consider the following example:
- Client A connects to the wss.
- Client B connects to the wss.
- Client A requests to lock storyline X.
- Client B requests to lock storyline Y.
- Server broadcasts to all clients that X is locked, along with its secret.
- Client A responds by setting X's secret in the store and stops listening for a response.
- Client B responds by setting X's secret in the store and stops listening for a response.
- Server broadcasts to all clients that Y is locked, along with its secret.
- Nothing happens since both clients already have a "secret".

We can see here that the result is dependent on the order of operations, and when you have multiple clients, as was the case in the testing party, chaos ensues!

The easiest fix was for the wss to respond only to client requesting the storyline lock/unlock and change a few order of operations. Unfortunately this does mean we lose out on the benefit stated above, but it does remove the possibility of any race conditions. In case we want to add it back, I have opted to comment it out instead of removing it altogether.

### Testing
Steps:
1. Open the dev site in several tabs (or partner up with someone on the team).
2. Attempt to load and save several different storylines simultaneously or as close to each other as possible.
3. Ensure everything is respectful.
4. Feel free to also verify that timeouts work.


IshavSohal (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/557)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mohsin-r)

spencerwahl (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/557)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mohsin-r)

yileifeng (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/557)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mohsin-r)

mohsin-r:
Closing this PR as this approach had issues. Opening a new one for the new implementation.

PR #577, opened 2025-03-20
Create admin upload component

IshavSohal:
### Related Item(s)
$426

### Changes
- Create component in the dashboard that allows admins to upload a zipped product file to the server
- Changed the layout of the dashboard buttons to use grid

### Notes
- Currently I'm using a modal for the upload component. Would it be better to create a separate page instead

### Testing
Steps:
1. Open the dashboard
2. Click the upload zip button to open the modal
3. Try submitting without any input. Note the toast error
4. Try submitting with just a uuid input. Notice the same error
5. Enter a unique UUID, and upload a product zip file (can be obtained from editor-main)
6. Press upload
7. Once it completes, check your local server to ensure the product is there with the UUID you entered
8. Try this again but with an existing UUID, and notice the toast error
9. Try dragging a non-zip file into the upload area. Notice the toast error
10. Try dragging multiple zip files to the upload component, and notice the toast info

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -61,7 +61,7 @@ app.use((req, res, next) => {
 // POST requests made to /upload will be handled here.
 app.route(ROUTE_PREFIX + '/upload/:id').post(function (req, res, next) {
     // Before any operation can be performed with the storyline, we need to ensure that the requester is the one who holds the lock for this storyline.
-    if (!verifySecret(req.params.id, req.headers.secret)) {
+    if (req.headers.secret !== 'N/A' && !verifySecret(req.params.id, req.headers.secret)) {
         res.status(400).send({
             status: 'Storyline was not locked or secret key corresponding to storyline lock incorrect/not provided.'
         });
--- a/src/components/home.vue
+++ b/src/components/home.vue
@@ -42,8 +42,8 @@
                 {{ $t('editor.landing.greeting') }} {{ userName }}!
             </div>
             <div class="mb-5 text-2xl font-semibold">{{ $t('editor.chooseOption') }}</div>
-            <div class="flex justify-center">
-                <div class="home-btn-container border border-gray-400 border-solid mr-5 flex-1 home-buttons">
+            <div class="button-grid">
+                <div class="home-btn-container border border-gray-400 border-solid home-buttons">
                     <router-link :to="{ name: 'metadataNew' }" class="flex justify-center h-full" target>
                         <button class="dashboard-option-button flex items-center text-xl font-bold px-2" tabindex="-1">
                             <svg
@@ -62,7 +62,7 @@
                         </button>
                     </router-link>
                 </div>
-                <div class="home-btn-container border border-gray-400 border-solid flex-1 home-buttons">
+                <div class="home-btn-container border border-gray-400 border-solid home-buttons">
                     <router-link :to="{ name: 'metadataExisting' }" class="flex justify-center h-full" target>
                         <button class="dashboard-option-button flex items-center text-xl font-bold" tabindex="-1">
                             <span class="pr-3">
@@ -103,7 +103,91 @@
                         </button>
                     </router-link>
                 </div>
+                <div
+                    class="home-btn-container border border-gray-400 border-solid home-buttons flex justify-center h-full"
+                    target
+                    @click="$vfm.open('admin-upload-modal')"
+                    v-if="profile.role !== 'Admin'"
+                >
+                    <button
+                        class="dashboard-option-button flex items-center justify-center text-xl font-bold w-full"
+                        tabindex="-1"
+                    >
+                        <span class="pr-3">
+                            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="55" height="55">
+                                <path
+                                    d="M20 6H12L10 4H4C2.9 4 2 4.9 2 6V18C2 19.1 2.9 20 4 20H20C21.1 20 22 19.1 22 18V8C22 6.9 21.1 6 20 6M20 18H16V16H14V18H4V8H14V10H16V8H20V18M16 12V10H18V12H16M14 12H16V14H14V12M18 16H16V14H18V16Z"
+                                />
+                            </svg>
+                        </span>
+                        <div style="font-size: calc(60% + 0.25vw)">{{ $t('editor.adminUpload') }}</div>
+                    </button>
+                </div>
             </div>
+
+            <vue-final-modal
+                modalId="admin-upload-modal"
+                content-class="edit-metadata-content max-h-full overflow-y-auto max-w-xl mx-4 p-7 bg-white border rounded-lg"
+                class="flex justify-center items-center"
+            >
+                <div class="p-5">
+                    <div class="text-xl text-center pb-8 font-bold">{{ $t('editor.adminUpload') }}</div>
+                    <div class="flex flex-col items-center pb-5">
+                        <label for="product-uuid">{{ $t('editor.uuid.new') }}: </label>
+                        <input
+                            type="text"
+                            name="uuid"
+                            id="product-uuid"
+                            class="border-solid border-2 w-5/6 p-2 mt-2 text-center"
+                            @change="productUuidChanged"
+                            placeholder="4523ae53-5d58-4a0e-8b01-54d9824871f5"
+                        />
+                    </div>
+                    <div
+                        class="flex flex-col items-center justify-center m-5 p-12 bg-blue-100 border-4 border-dashed border-blue-300"
+                        :class="{ dragging: isDragging }"
+                        @dragover.prevent="() => (dragging = true)"
+                        @dragleave.prevent="() => (dragging = false)"
+                        @drop.prevent="productZipDropped($event)"
+                    >
+                        <label class="cursor-pointer p-4">
+                            <span>{{ $t('editor.image.label.drag', { fileType: 'zip file' }) }}</span>
+                            {{ $t('editor.label.or') }}
+                            <span class="text-blue-700 font-bold">{{ $t('editor.label.browse') }}</span>
+                            {{ $t('editor.label.upload') }}
+                            <input
+                                type="file"
+                                name="zip"
+                                id="product-zip-file"
+                                @change="productZipUploaded"
+                                multiple="false"
+                                accept="application/x-zip-compressed"
+                                class="cursor-pointer"
+                            />
+                        </label>
+                        <div>{{ $t('editor.adminUpload.fileName') }}: {{ productZipName || 'N/A' }}</div>
+                    </div>
+
+                    <div class="flex justify-end">
+                        <div class="inline-flex align-middle mr-2" v-if="loading">
+                            <spinner size="24px" color="#009cd1" class="mx-2 my-auto"></spinner>
+                        </div>
+                        <button
+                            type="submit"
+                            @click="uploadZip"
+                            class="editor-button editor-forms-button bg-black text-white hover:bg-gray-80 mr-2"
+                        >
+                            {{ $t('editor.video.label.upload') }}
+                        </button>
+                        <button
+                            @click="$vfm.close('admin-upload-modal')"
+                            class="editor-button editor-forms-button border border-gray-300"
+                        >
+                            {{ $t('editor.cancel') }}
+                        </button>
+                    </div>
+                </div>
+            </vue-final-modal>
 
             <h2 class="pt-8 pb-5 text-2xl font-semibold">{{ $t('editor.previousProducts') }}</h2>
             <table class="shadow-lg bg-white w-full pr-0 mr-0">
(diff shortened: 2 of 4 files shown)


IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`src/components/home.vue` line 110 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OLoAVhI568csw5ap56l:-OLoAVhI568csw5ap56m:b-8bfw0b) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/670dae3e528646b847f3735a83e0d182f5a7d299/src/components/home.vue#L110)):*
> ```Vue
>                     target
>                     @click="$vfm.open('admin-upload-modal')"
>                     v-if="profile.role !== 'Admin'"
> ```

Blocker

mohsin-r (changes requested):
Would be nice from a convenience perspective if we could select a folder and it automatically gets some zipped up and uploaded instead of having to do the zipping part ourself, but this is not urgent or too too important.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-:-OM7sYHn49N84vS-a0bl:bdb14k8)* status: all files reviewed, 2 unresolved discussions

___
*[`server/index.js` line 64 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OM7s6PsDnQEWyb9WLrw:-OM7s6Pt-RVSh5L3c8oc:b-7srjx8) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/670dae3e528646b847f3735a83e0d182f5a7d299/server/index.js#L64)):*
> ```JavaScript
> app.route(ROUTE_PREFIX + '/upload/:id').post(function (req, res, next) {
>     // Before any operation can be performed with the storyline, we need to ensure that the requester is the one who holds the lock for this storyline.
>     if (req.headers.secret !== 'N/A' && !verifySecret(req.params.id, req.headers.secret)) {
> ```

Lol nice one, but we need to lock the UUID, upload the product, and then unlock it on upload donethanks, or we risk overwriting someone elses's work.

IshavSohal:
I've added some todo's for this, will work on it once this PR is merged

szczz (commented):
Really well done! Couple things

1. Ideally we should have the duplicate UUID check occur inside the modal.

2. I uploaded a zip with the name of an existing folder (Don't think this is relevant) and successfully uploaded it. I can see the folder in /public with the UUID I provided in the input, but when I go to load it I get the successfully loaded storyline toast, but the metadata component isn't rendering.



> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-:-OMCZmyL66i-4_JHwPrF:b-ybueph)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @mohsin-r)

IshavSohal (commented):
> Ideally we should have the duplicate UUID check occur inside the modal.

Donethanks. Now an error message will appear under the UUID input whenever the UUID entered already exists or contains invalid characters. The `Upload` button will also be disabled in such a scenario. There are some formatting issues with the modal when the error message appears, so I'll try to fix those once this PR is merged. 

> I uploaded a zip with the name of an existing folder (Don't think this is relevant) and successfully uploaded it. I can see the folder in /public with the UUID I provided in the input, but when I go to load it I get the successfully loaded storyline toast, but the metadata component isn't rendering.

Donethanks. I created a new endpoint in the server (`/adminRename`) that completes the renaming process for the uploaded product. The endpoint will change the names of the json config files to use the UUID provided by the user, as well as replace every instance of the old UUID (within a path) with the new UUID.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-:-OMIy-9m9e5kpU5X8QK8:b1ujupl)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @mohsin-r)

mohsin-r (changes requested):
> Reviewed 3 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577)* status: 4 of 5 files reviewed, 3 unresolved discussions

___
*[`server/index.js` line 439 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OMRyeZ49UPmkHhZvIgq:-OMRyeZ55m69s5rheiah:b-62gz4t) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/c27096e61555ebbd496275a122a14557966f1e2b/server/index.js#L439)):*
> ```JavaScript
> });
> 
> app.route(ROUTE_PREFIX + '/adminRename').post(function (req, res) {
> ```

This is good, but we want to protect this route by checking for a lock for the new UUID first, similar to how it is implemented in the other routes. Then, the frontend should pass in the correct secret. 

Also, a bit more a petty grouse but generally I think route naming uses  kebab case so `/admin-rename` instead of `/adminRename`
___
*[`src/lang/lang.csv` line 44 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OMRzZoC7Co9DaW1eAoe:-OMRzZoC7Co9DaW1eAof:bvwgi3d) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/c27096e61555ebbd496275a122a14557966f1e2b/src/lang/lang.csv#L44)):*
> ```CSV
> editor.adminUpload.oneFileOnly,You are only allowed to upload one file. The first of the files you uploaded has been chosen.,1,Vous n'êtes autorisé à télécharger qu'un seul fichier. Le premier des fichiers que vous avez téléchargés a été choisi.,0
> editor.adminUpload.fileName,Name of file uploaded,1,Nom du fichier téléchargé,0
> editor.adminUpload.lockFail,Failed to lock {uuid},1,Échec du verrouillage de {uuid},0
> ```

Ideally we don't really want to use the word lock for a user-displayed message as its a more technical term. You can probably steal one of the other strings in there that says something like "This product is being accessed by another user..."

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on @mohsin-r)

___
*[`server/index.js` line 439 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OMRyeZ49UPmkHhZvIgq:-OMT3Lsq0Yb7B74RxaXR:behjw3g) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/c27096e61555ebbd496275a122a14557966f1e2b/server/index.js#L439)):*

This is good, but we want to protect this route by checking for a lock for the new UUID first, similar to how it is implemented in the other routes. Then, the frontend should pass in the correct secret. 

Also, a bit more a petty grouse but generally I think route naming uses  kebab case so `/admin-rename` instead of `/adminRename`
</blockquote></details>

Good catch, donethanks
___
*[`src/lang/lang.csv` line 44 at r3](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OMRzZoC7Co9DaW1eAoe:-OMT3PUw7UQPVdY2NdHC:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/c27096e61555ebbd496275a122a14557966f1e2b/src/lang/lang.csv#L44)):*

Ideally we don't really want to use the word lock for a user-displayed message as its a more technical term. You can probably steal one of the other strings in there that says something like "This product is being accessed by another user..."
</blockquote></details>

Donethanks

szczz (changes requested):
Getting a failed to rename error after uploading a zip. I can see the folder in /public and when I try to load it I get the successfully loaded storyline toast but still nothing renders in the metadata component. @yileifeng is also experiencing this issue

> Reviewed 1 of 4 files at r1, 1 of 4 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-:-ONa2hu5E-Acaw3uwkM_:bawu37u)* status: all files reviewed, 1 unresolved discussion (waiting on @mohsin-r)

yileifeng (commented):
The reason I was getting the issue was because I zipped a Storyline from its root folder manually instead of zipping all its contents (assets, ramp-config, styles, config files). Trying to upload the zipped Storyline using this method uploaded the folder onto the server with an additional directory level: `[UUID]/[UUID]/[content]`. Uploading a ZIP folder with the proper structure works correctly (e.g. using an exported product from the main editor). 

This brings up a thought: since this functionality is primarily designed for uploading Storyline products that were not made through RESPECT, do we need to check for restrictions in the uploaded file structure since any type of ZIP file can be uploaded currently?

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-:-ONuUkjW9f4YVCVVKoZ-:b-haajno)* status: all files reviewed, 1 unresolved discussion (waiting on @mohsin-r and @szczz)

IshavSohal (commented):
This makes sense to me. Since there isn't much of a point in uploading a product already available to the editor, manually zipping a product folder is probably the way this upload component would be used most of the time. The only issue is that the `/upload` endpoint expects the zipped file to be a product folder upon unzipping, rather than a product folder nested in another folder. So, if we wanted to accept a zip file that unzips to a product folder nested in another folder, we would need to update the `/upload` endpoint, or create a new endpoint.

In case we want to stick with the current implementation, I added an info icon to the upload component, which provides steps on how a product zip file can be obtained. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-:-ONuvQGhAJL-rBtJd0-Y:b9b5jao)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @mohsin-r and @szczz)

yileifeng (changes requested):
I'm good with the current implementation + the extra info text :+1:

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-:-OOSUAJiAvpkMpT68yM9:bef61zu)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @mohsin-r and @szczz)

___
*[`src/lang/lang.csv` line 44 at r6](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OOSTvesCGdr5gYIi_y7:-OOSTvesCGdr5gYIi_y8:b-wt0edb) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/071e0db70720269d854ee71db595717be1c9eadf/src/lang/lang.csv#L44)):*
> ```CSV
> editor.adminUpload.oneFileOnly,You are only allowed to upload one file. The first of the files you uploaded has been chosen.,1,Vous n'êtes autorisé à télécharger qu'un seul fichier. Le premier des fichiers que vous avez téléchargés a été choisi.,0
> editor.adminUpload.fileName,Name of file uploaded,1,Nom du fichier téléchargé,0
> editor.adminUpload.zipFileInfo,"A zipped product folder can be obtained by navigating to a product folder, highlighting all of it's contents, right clicking and then compressing to a zip file. Alternatively, you can use the export component within editor-main",0,"Un dossier de produit zippé peut être obtenu en naviguant vers un dossier de produit, en mettant en surbrillance tout son contenu, en cliquant avec le bouton droit de la souris, puis en le compressant dans un fichier zip. Sinon, vous pouvez utiliser le composant d'exportation dans editor-main",0
> ```

I think we can drop the "Alternatively, you can use the export component within editor-main" to keep it concise since we probably would not be uploading the product again if it already existed in the editor

IshavSohal (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @mohsin-r, @szczz, and @yileifeng)

___
*[`src/lang/lang.csv` line 44 at r6](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577#-OOSTvesCGdr5gYIi_y7:-OOSbZOC7_3L4xiIAPBO:bah2568) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/071e0db70720269d854ee71db595717be1c9eadf/src/lang/lang.csv#L44)):*

I think we can drop the "Alternatively, you can use the export component within editor-main" to keep it concise since we probably would not be uploading the product again if it already existed in the editor
</blockquote></details>

Donethanks

szczz (approved):
> Reviewed 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/577)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IshavSohal)

dan-bowerman:
From 2025-04-29 Testing Party:

1. Upload a Storylines zip (e.g. [08d9bd86-6535-4daa-8284-c5ade064982c.zip](https://github.com/user-attachments/files/19961043/08d9bd86-6535-4daa-8284-c5ade064982c.zip))
2. Name it something (e.g. `testytesty`) 
3. Try to load it. There's no Next button to do anything further.

yileifeng:
^Confirming same behavior above, even without web socket errors uploaded product was not loading. However, it was definitely working on the local server when tested.

szczz:
In addition to this, it doesn't seem like the local storyline store is being updated with the newly upload product

IshavSohal:
I believe this is due to the garbage collection code that was recently merged in. With that code, we are obtaining and stringifying the configs, so we can search them for embedded file sources. However, when a zip file is uploaded, the UUID provided by the user, and thus the UUID provided to the `/upload` endpoint, may not be the same as the name used for the config files. This results in an error on [this](https://github.com/ramp4-pcar4/storylines-editor/blob/5fcf2420f5bd7a06ba93207ff7e71a68c69cf774/server/index.js#L128) line in the server.

In this case, we can just prevent the garbage collection code from running. Once the admin uploaded product is appropriately renamed (using the `admin-rename` endpoint), garbage collection should work fine whenever the `/upload` endpoint is called.

Issue #559, opened 2025-02-25
Cancel Canada.ca WET styling within editor component


gordlin:
## Description

Currently, when you go onto the Canada.ca template, the WET styles in the page template conflict with some of the storyline editor's internal stylings.

### Is your feature request related to a problem?

Luckily, this doesn't cause major issues, but various styling problems do occur:

- Various buttons get blue text/underlines when hovered
- Stuff gets un-aligned
- Problems 

These problems will only happen on the Canada.ca template and not the standalone editor, creating a mismatch which can make it more annoying to develop. Best to find a way to reunify them.

## Proposed solution

Find a way to cancel out the Canada.ca template styles within the storylines editor component.

## Alternatives considered

There are various potential methods for implementing this:

- The most promising seems to be using a **shadow DOM** to completely separate the component from surrounding page styles. More info [here](https://matthewjamestaylor.com/style-blocker). 
- Another potential method is using an **iframe** to embed the editor. 

## Related Issues

Discussed in https://github.com/ramp4-pcar4/storylines-editor/pull/530#pullrequestreview-2631350945

PR #584, opened 2025-04-01
Override WET styles within RESPECT instance

IshavSohal:
### Related Item(s)
#559

### Changes
- For the Canada.ca template, create a copy of each Vue component style (i.e. any <style> element found in the document) and wrap it with the `#app` selector, so that it has higher specificity than any WET styles being applied to the RESPECT instance
- A copy is created because some styles within a stylesheet are intended for the root element, while others are intended for the RESPECT instance
- So any styles meant for the root will be applied as needed, and any styles intended for the RESPECT instance will override WET styles

### Notes
- I was able to get the RESPECT instance within a shadow dom without too many issues. However, the `json-editor` component used for the advanced editor fails to render. This is due to the fact that the component accesses the page using `document`, which will not work when the RESPECT instance is within a shadow dom. We would have to create a local fork of the `json-editor` repo, and add the ability to provide it's parent element in order to make this work
- Getting an iframe to work is quite complicated, and leads to performance issues.

### Testing
Steps:
1. Open the dashboard
2. Hover over the buttons, and notice that the text doesn't get a blue underline anymore
3. Load a product into `editor-main`
4. Go to a slide and toggle one of the slide controls
5. Notice that the slide control gets a blue highlight
6. Make sure there are no other styling discrepancies between RESPECT being inside and outside of the Canada.ca template


james-rae (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/584)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IshavSohal)

Issue #426, opened 2024-11-19
Create "Admin" only upload component 


szczz:
## Description

Discussed in the 11/19/2024 Platform meeting.

We want to introduce a component that allows for uploading Storyline zips to a server env. Copying folders from local machines onto the virtual folders on the server creates issues with the git integration, so this new component will allow us to avoid the grousing.

This component should only be accessible to users that have the "Admin" role. Roles should already be accessible in the userStore, however, we haven't used them elsewhere in RESPECT yet. The component should have a simple interface that allows the user to drag and drop a zip or select it from the file system. It should then create a new folder on the server and extract the contents of the zip into this folder. **Note**: The folder itself must be created by the server, so we can't just extract the zip directly onto the server. 

We will also need to validate that the server does not already contain a Storyline with the same UUID as the uploaded folder. We'll likely want to add a UUID input field in this component in case there is a naming conflict. Additionally, we likely want a link to this component added to the dashboard which should only be visible to users with the "Admin" role. 

Once deployed to dev please reach out to me to have your role set to "Admin" for testing.

PR #595, opened 2025-05-06
Enhance admin upload

IshavSohal:
### Related Item(s)
#426

### Changes
- Allow for the uploading of product folders as is, without having to zip them
   - Works for both file selection and drag/drop
   - Had to use the `file-selector` 3rd party library for drag/drop
- Improved modal formatting/layout for smaller screen sizes
- Other styling improvements in dashboard

### Notes
- Assuming the folder upload is working correctly, should we still include the zip upload?

### Testing
Steps:
1. Ensure you have the server changes locally
2. Open the admin upload modal
3. Upload a product folder using the second input box
4. Provide a UUID and upload. Ensure that the product is correctly uploaded into your local server
5. Load the product into the editor and ensure that there are no issues
6. Go back to the admin upload modal and repeat steps 2-4 by dragging/dropping a product folder
7. Try uploading a random folder (one that doesn't adhere to the storylines product structure), and observe an appropriate error message
8. Try dragging/dropping a regular file, and observe an appropriate error message

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
                 "deep-object-diff": "^1.1.9",
                 "dompurify": "^3.2.6",
                 "file-saver": "^2.0.5",
+                "file-selector": "^2.1.2",
                 "highcharts": "^9.3.2",
                 "highcharts-accessible-configuration-kit": "^1.15.0",
                 "highcharts-vue": "^1.4.3",
@@ -5907,6 +5908,24 @@
             "version": "2.0.5",
             "resolved": "https://registry.npmjs.org/file-saver/-/file-saver-2.0.5.tgz",
             "integrity": "sha512-P9bmyZ3h/PRG+Nzga+rbdI4OEpNDzAVyy74uVO9ATgzLK6VtAsYybF/+TOCvrc0MO793d6+42lLyZTw7/ArVzA=="
+        },
+        "node_modules/file-selector": {
+            "version": "2.1.2",
+            "resolved": "https://registry.npmjs.org/file-selector/-/file-selector-2.1.2.tgz",
+            "integrity": "sha512-QgXo+mXTe8ljeqUFaX3QVHc5osSItJ/Km+xpocx0aSqWGMSCf6qYs/VnzZgS864Pjn5iceMRFigeAV7AfTlaig==",
+            "license": "MIT",
+            "dependencies": {
+                "tslib": "^2.7.0"
+            },
+            "engines": {
+                "node": ">= 12"
+            }
+        },
+        "node_modules/file-selector/node_modules/tslib": {
+            "version": "2.8.1",
+            "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.8.1.tgz",
+            "integrity": "sha512-oJFu94HQb+KVduSUQL7wnpmqnfmLsOA/nAh6b6EH0wCEoK0/mPeXU6c3wKDV83MkOuHPRHtSXKKU99IBazS/2w==",
+            "license": "0BSD"
         },
         "node_modules/fill-range": {
             "version": "7.1.1",
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
         "deep-object-diff": "^1.1.9",
         "dompurify": "^3.2.6",
         "file-saver": "^2.0.5",
+        "file-selector": "^2.1.2",
         "highcharts": "^9.3.2",
         "highcharts-accessible-configuration-kit": "^1.15.0",
         "highcharts-vue": "^1.4.3",
--- a/server/index.js
+++ b/server/index.js
@@ -546,15 +546,15 @@ app.route(ROUTE_PREFIX + '/admin-rename').post(function (req, res) {
                     let configEn = JSON.stringify(JSON.parse(data), null, 4);
                     configEn = configEn.replaceAll(`${configUuid}/`, `${newUuid}/`);
 
-                    // Create the new en config file
-                    fs.writeFile(PRODUCT_PATH + `/${newUuid}_en.json`, configEn, 'utf8', async (err) => {
+                    // Remove the old en config from the product
+                    fs.unlink(PRODUCT_PATH + `/${configUuid}_en.json`, (err) => {
                         if (err) {
                             res.status(500).send({ status: 'Internal Server Error' });
                             logger('WARNING', 'Error occured while renaming a Storylines product.' + err);
                             return;
                         }
-                        // Remove the old en config from the product
-                        fs.unlink(PRODUCT_PATH + `/${configUuid}_en.json`, (err) => {
+                        // Create the new en config file
+                        fs.writeFile(PRODUCT_PATH + `/${newUuid}_en.json`, configEn, 'utf8', async (err) => {
                             if (err) {
                                 res.status(500).send({ status: 'Internal Server Error' });
                                 logger('WARNING', 'Error occured while renaming a Storylines product.' + err);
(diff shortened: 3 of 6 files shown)


IshavSohal:
*[`src/components/home.vue` line 110 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-OPb6DIk1YHZdJltoAr3:-OPb6DIk1YHZdJltoAr4:b-8bfw0b) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/294cdd0e5e7c86e8d4b615880551423a99a79bda/src/components/home.vue#L110)):*
> ```Vue
>                     target
>                     @click="$vfm.open('admin-upload-modal')"
>                     v-if="profile.role !== 'Admin'"
> ```

Blocker

james-rae (approved):
> Reviewed 4 of 6 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595)* status: all files reviewed, 1 unresolved discussion

RyanCoulsonCA (changes requested):
It seems like if you upload a product ZIP or folder and keep the UUID the same, the configuration files get deleted.

Steps to replicate:
1. Ensure you don't have a product in your server folder with the UUID of the product you're going to upload.
2. Upload a new product from a different folder. It can be a ZIP or just the folder.
3. Enter the product UUID (keep it the same as the folder name).
4. Upload.
5. Notice that the files are uploaded, and then the configuration files get removed.

Demo:


> Reviewed 3 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-:-ORWT1tf5CLeiMih-weH:b-7scyii)* status: all files reviewed, 2 unresolved discussions (waiting on @james-rae)

___
*[`src/components/home.vue` line 133 at r2](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-ORWhL6N3ilRakI-AUPB:-ORWhL6N3ilRakI-AUPC:bb5g2fl) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/c63821a1abaaf1d2f7d89d330829278c1ac773e2/src/components/home.vue#L133)):*
> ```Vue
>                 class="flex justify-center items-center"
>             >
>                 <div class="modal-contents">
> ```

This doesn't have to be done now, just leaving this here as a note for a future refactor (it needs to be done in a bunch of places throughout the app, not just here). 

We should break this modal and the functions it requires out into it's own component. It'd help to keep this `home` component simple.

IshavSohal (commented):
Donethanks, had to switch the order of writing the new file and deleting the old file in the `admin-rename` endpoint. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-:-ORXIYjo0OtSA0SkYkXU:bwtzwwf)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @james-rae and @RyanCoulsonCA)

___
*[`src/components/home.vue` line 133 at r2](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-ORWhL6N3ilRakI-AUPB:-ORXIgLo9h6iUxJIke21:b-qma6kb) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/c63821a1abaaf1d2f7d89d330829278c1ac773e2/src/components/home.vue#L133)):*

This doesn't have to be done now, just leaving this here as a note for a future refactor (it needs to be done in a bunch of places throughout the app, not just here). 

We should break this modal and the functions it requires out into it's own component. It'd help to keep this `home` component simple.
</blockquote></details>

Good point, I'll open an issue for this. Also the `/components` folder should probably be organized into subfolders, similar to RAMP.

szczz (changes requested):
If I copy a folder named CopyParty and rename it locally to CopyParty2 the frontend grouses at me with "Uploaded folder doesn't match product structure". This is because the config files don't have the same name as the root folder. However, we're forced to provide a UUID anyways, so it seem like if we're going to enforce the folder name matching the configs we should probably do it after we press upload. The error message should also be a bit more descriptive. 

Edit: actually... if we're giving the ability to rename the folder, then we'll likely just want to rename the configs to match?

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-:-OSK7y0C7XCe3TYSvJCd:b-bzy6a9)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @james-rae and @RyanCoulsonCA)

IshavSohal:
Yep we're renaming the config files within the `/admin-upload` endpoint to use the user provided UUID. Within that endpoint, we are using the config file names to obtain the old UUID used in the file paths throughout the config, so that we can replace those old UUIDs with the new, user provided UUID. This is why there is a check to ensure that the folder name (which corresponds to the old UUID) matches with the config file names.

Should we be handling the case where the folder name doesn't match the config file names? This would only occur if the user manually updated the folder name or config file names, but by using the editor itself it shouldn't be possible. If this case is worth handling, I can modify the `/admin-upload` endpoint by passing in the folder name as the old uuid (would be an optional param), rather than relying on the config file names. For a zip file upload, however, we would need to rely on the config file names.

szczz (approved):
Yea I could see scenarios where an admin needs to duplicate/fork a product. They might download the zip of product xyz and rename it to xyz-fork or something so its easily identifiable. Gonna add a meeting topic tag to this. 

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-:-OSt8gA6BE8EyVGmp-LI:b1vqkr6)* status: all files reviewed, 2 unresolved discussions (waiting on @james-rae and @RyanCoulsonCA)

RyanCoulsonCA (approved):
> Reviewed 3 of 6 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595)* status: all files reviewed, 1 unresolved discussion (waiting on @james-rae and @szczz)

IshavSohal (commented):
Now there is a single area for dragging/dropping both zip files and product folders. However I had to include both file input elements, since it isn't possible for a single file input element to accept both product folders and zip files. Things should be working the same as before, but maybe a quick round of testing wouldn't hurt. 

I've also put the info content for both upload options into a single info icon. In case this is too long, I've also put the info content as a tooltip on the labels corresponding to the file inputs. Lmk which one we should keep.

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595#-:-OVSh0CuADeuFQ7GN9o4:b-o61x2d)* status: 1 of 7 files reviewed, 1 unresolved discussion (waiting on @james-rae, @RyanCoulsonCA, and @szczz)

james-rae (approved):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595)* status: all files reviewed, 1 unresolved discussion (waiting on @RyanCoulsonCA and @szczz)

spencerwahl (approved):
> Reviewed 1 of 6 files at r3, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595)* status: all files reviewed, 1 unresolved discussion (waiting on @RyanCoulsonCA and @szczz)

spencerwahl (approved):
> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/595)* status: all files reviewed, 1 unresolved discussion (waiting on @james-rae, @RyanCoulsonCA, @szczz, and @yileifeng)

Issue #616, opened 2025-05-30
Products with capitalized letters in UUID cannot be loaded using lowercase


IshavSohal:
## Description

If a product's uuid contains capital letters, attempting to load in the product using an all-lowercase uuid will result in a success toast, but no metadata-editor

## Steps to reproduce

1. Create a new product with some upper case letters in the uuid
2. Save the product
3. Try to load the product in using the uuid exactly how it is
4. Ensure that this works
5. Now reload and try again with the capital letters replaced with lowercase
6. Notice the success message, but no metadata editor

## Expected behavior

Either we should be displaying an error message, or we should load the product in correctly

## Possible solution

Within `loadConfig()`, we are attempting to obtain the config files using the `uuid` that the user typed in, which would not work if  `uuid` is all-lowercase and the uuid on the server contains capital letters

PR #618, opened 2025-06-02
Correct uuid with incorrect letter case

IshavSohal:
### Related Item(s)
#616

### Changes
- When a uuid is entered with the incorrect letter case, ensure that it is correct prior to attempting to read the config files within `configFileStructure`

### Notes
- Here I'm assuming that the names of the files within `configFileStructure` are file paths relative to the product folder on the server. So, the json files that are sitting directly in the product folder (and not in a sub folder) would be the config files
- This is the case locally, but let me now if this is not the case on the dev server 

### Testing
Steps:
1. Load in a product by entering its uuid manually, but with incorrect letter case (lowercase instead of uppercase, or vice versa)
2. Ensure that the metadata editor and Next button loads in
3. Click Next and ensure that the product is loaded into editor-main
4. Make some changes and save
5. Ensure that changes are saved successfully

### DIFF
--- a/src/components/metadata-editor.vue
+++ b/src/components/metadata-editor.vue
@@ -1990,7 +1990,26 @@ export default class MetadataEditorV extends Vue {
                 }
             });
         }
+        this.correctUuid();
         return this.loadConfig();
+    }
+
+    /**
+     * Ensure that `uuid` is a case-sensitive match with the product's uuid on the server
+     */
+    correctUuid(): void {
+        const configFileNames = Object.keys(this.configFileStructure.zip.files).filter(
+            (key) => key.includes('.json') && !key.includes('/')
+        );
+        if (configFileNames.length > 0) {
+            const productUuid = configFileNames[0].split('_')[0]; // Case-sensitive UUID of product on server
+
+            // Only update uuid if its the same as on the server, just with incorrect letter case
+            if (productUuid.toLowerCase() === this.uuid.toLowerCase()) {
+                this.uuid = productUuid;
+                this.configFileStructure.uuid = productUuid;
+            }
+        }
     }
 
     /**


james-rae (approved):
question: is case sensitivity also considered when making a new storyline? E.g. If I have storyline `dogs` already saved, then create a new storyline with uuid `DOGs`, will it run the "you're replacing a storyline?" or will it make a new one? Or will it think its new, and then silently steamroll the old one due to folder names being the same?

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/618#-:-OS0Lbez1rND8F2JCjbM:bfmmbmy)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IshavSohal)

szczz (approved):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/618)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IshavSohal)

IshavSohal:
From the looks of it, the check would be case-insensitive. So if you tried to create a new `DOGs` storyline, you would get an error saying 'UUID already exists. Please choose a different one', and you would be unable to create the new product.

RyanCoulsonCA (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/618)* status: all files reviewed, 1 unresolved discussion (waiting on @IshavSohal)

___
*[`src/components/metadata-editor.vue` line 2005 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/618#-OS4yBzc4vymdiSzaztb:-OS4yBzc4vymdiSzaztc:bu8zpt7) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/5c48dc7da9f8bdb623e2cc4b9ed2fa45cc918865/src/components/metadata-editor.vue#L2005)):*
> ```Vue
>         );
>         if (configFileNames.length > 0) {
>             const productUuid = configFileNames[0].split('_')[0]; // Case-sensitive UUID of product on server
> ```

The issue is still occurring for products that have UUIDs which contain underscores (WQMS is `WQMS_GreatLakes-GrandLacs_DQME` for example). Modifying this line to take everything before the *last* underscore in the file name should fix this.

IshavSohal:
Just realized that the locks are case-sensitive, so it's possible to lock the same product twice. Will try to find a workaround.

Issue #549, opened 2025-02-18
Console spamming on UUID checks


james-rae:
## Description

It looks like the landing page is checking if a uuid in the box is valid by pinging the server.
So when you type a new custom uuid, your generating an invalid ping on every keypress.

That in itself is fine (since this isn't a frequent thing, so not like server is getting mashed), but each fail writes to the console.

## Proposed solution

If possible, wrap this test in something that handles the error before it hits the console.

## Additional context

![Image](https://github.com/user-attachments/assets/7d0829ca-cec4-4145-b36e-7c81f3ee3754)

PR #694, opened 2025-07-15
Prevent console spamming when entering new uuid

rrjk76:
### Related Item(s)
Issue #549 

### Changes
- Extracted the `uuid` availability check into a new `fetchUuidCheck` function. 
- Added `debounce` (1500ms) to `fetchUuidCheck` so that the fetch (and the post) only execute when the user has stopped typing.
-  Moved the `uuid` validation logic (checking illegal characters or blank input) into `isUuidInvalid` so that it is reusable.

### Notes
- This doesn't suppress the 404 warnings in the browser's console entirely, and if the user types slowly, more fetches will execute and result in more console log warnings.

- A possible workaround that _could_ get rid of all 404 warnings would be to register a Service Watcher (e.g., `no404-sw.js`)  to intercept 404s and rewrite them as 202s (for example), which hides the warnings in the console.
  - I tested this out and it seems to work. The original response status (404) can still be accessed from the custom header `X-Original-Status`, so custom warnings are still possible. 
  - That said, because this approach affects all 404s across the app, I'm not sure if it's a good idea to remove all network warnings, so for now I've just delayed each fetch. 

### Testing
Steps:
1. Create or load an existing product. 
2. Open the browser's devtools console.
3. Edit the `uuid`. 
4. Notice that no 404 errors are logged on every keypress while typing.
5. 404 warnings appear only after you are done typing (or after a pause).

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/694)

### DIFF
--- a/src/components/metadata/metadata-editor.vue
+++ b/src/components/metadata/metadata-editor.vue
@@ -32,7 +32,9 @@
                         v-if="!currentRoute.includes('index-ca')"
                         :to="{
                             name: editExisting ? 'metadataExisting' : 'metadataNew',
-                            params: { lang: currLang === 'en' ? 'fr' : 'en', uid: uuid }
+                            params: editExisting
+                                ? { lang: currLang === 'en' ? 'fr' : 'en', uid: uuid }
+                                : { lang: currLang === 'en' ? 'fr' : 'en' }
                         }"
                     >
                         <div class="respected-standard-link-button">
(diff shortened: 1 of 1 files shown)


spencerwahl (approved):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/694)* status: all files reviewed, 1 unresolved discussion (waiting on @james-rae)

___
*[`src/components/metadata/metadata-editor.vue` line 2175 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/694#-OVSZyTz9ILX8gnr3AZG:-OVSeJAtAFwbz_A0BXNL:bxhzvkf) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/559578384a6b0bbca61124d12fdc43b85bd2ff2a/src/components/metadata/metadata-editor.vue#L2175)):*
<details><summary><i>Previously, james-rae (James Rae) wrote…</i></summary><blockquote>

curious if you played with the debounce time. I can't get the demo to work to try it out; waiting 1.5 seconds "feels" like a user would notice the delay.  Just wondering if lower values were worse.
Was 300 before but also a `throttle` which plays a bit different.
</blockquote></details>

+1 to trying a smaller delay

PR #710, opened 2025-07-21
Fix edit button for charts

rrjk76:
### Related Item(s)
Issue [#111](https://github.com/ramp4-pcar4/highcharts-accessible-configuration-kit/issues/111#issue-3236935367) 

### Changes
- Edit button now works for all charts. 
- If the `highchartsJson` file is not found in the zip and the title is present, a new chart JSON file is generated and added to the configuration structure. 

### Notes
* The cause for the original issue was the provided `src`  (should have been `testytesterson/charts/fr/...` instead of `testytesterson/charts/en/...`).
  * To address this, I updated the `chartSrc` value when fetching the chart configs. 

* For other slides (7 & 9), clicking `Edit` did nothing because `highchartsJson` was false, and hence the chart JSON wasn’t loaded or created, so `openEditor` had no data to open.

### Testing
Steps:
1. Load the `testytesterson` product. 
2. Go to slides 5 (French panel), 7, or 9 (these were the charts where the edit button wasn't working originally). 
3. Press `Edit`, and the editor modal should open as expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/710)

### DIFF
--- a/src/components/panel-editors/chart-editor.vue
+++ b/src/components/panel-editors/chart-editor.vue
@@ -188,16 +188,7 @@ export default class ChartEditorV extends Vue {
 
         // fetch single existing chart config from ZIP
         if (this.panel.type === PanelType.Chart && this.panel.src) {
-            const chartSrc = `${this.panel.src.substring(this.panel.src.indexOf('/') + 1)}`;
-            const highchartsJson = this.productStore.configFileStructure.zip.file(chartSrc);
-            if (highchartsJson) {
-                highchartsJson.async('string').then((res: string) => {
-                    this.highchartsChartConfigs.push(JSON.parse(res));
-                });
-            }
-
-            this.extractStorylinesChartConfig(this.panel as ChartPanel);
-            this.chartIdx += 1;
+            this.fetchChartConfig(this.panel, 0, this.panel.config?.title?.text);
         }
 
         if (this.centerSlide && this.dynamicSelected) {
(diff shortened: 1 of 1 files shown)


yileifeng (changes requested):
There is another breaking case in which the Storylines config name does not match the highcharts config title, in which cases the open editor will fail to open since `this.highchartsChartConfigs.findIndex((cfg) \=> cfg.title.text \=== name)` returns -1. This case exists on the 0000 product with the `hybridChartConfig`. 

These two name properties should always be aligned when using the new highcharts editor, but older products might run into a mismatch (main cause is saved file name does not match chart title). One potential solution might be to add a chart title parameter to the `extractStorylinesChartConfig` function to verify that the name set in that function matches. 

> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/710#-:-OVhycgvEVw3VzsopU15:b-4ehig9)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/components/panel-editors/chart-editor.vue` line 191 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/710#-OVhyESs8itTDeuWD2XO:-OVhyESs8itTDeuWD2XP:bqn0rye) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/35f19894748952e7780e45c18f881a3fd84cfa33/src/components/panel-editors/chart-editor.vue#L191)):*
> ```Vue
>         // fetch single existing chart config from ZIP
>         if (this.panel.type === PanelType.Chart && this.panel.src) {
>             const chartSrc = this.panel.config?.title?.text
> ```

Maybe we can move this code block extracting the highcharts config into its own helper function since its repetitive with the `PanelType.Slideshow` case below

yileifeng (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/710)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @rrjk76)

szczz (approved):
So the edit button does open the charts now, however, I'm still seeing some issues with the first chart on slide 5 FR. The preview shows nothing plotted, but when I press edit I see plotted data and nothing in the data section. I'm guessing this will be fixed in another PR, so I won't block.




 

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/710#-:-OW0r8VW99sa9JPFTDZG:b8i9a2y)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @rrjk76)

IshavSohal (approved):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/710)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @rrjk76)

gordlin (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/710)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @rrjk76)

Issue #713, opened 2025-07-23
Allow "Admin" or "Dev" user roles to bypass script removal in text panels.


szczz:
## Description

#685 introduced the sanitation of text panels of common tags, however, it was mentioned that this feature might be better implemented on a role basis. Admins, or Dev team members should perhaps have the ability to bypass this restriction as it may be required for custom stroylines. 

I'll add the meeting topic tag for now as we should probably discuss this further before implementing it.

szczz:
Let proceed with adding a bypass to this functionality for Admin roles.

PR #737, opened 2025-08-08
add content sanitization bypass for admin role

RyanCoulsonCA:
### Related Item(s)
#713 

### Changes
- Adds a `sanitize` button to the text editor toolbar for users with the "admin" role, which allows admins to bypass content sanitization when working on a product.

### Notes
The "sanitize" selection gets assigned every time the text panel is mounted, which means that admins who want to sanitize content will need to re-enable it every time they open a text panel. This is pretty annoying, so maybe it's worth storing their preference in the user store for convenience.

### Testing
Steps:
1. Open the demo link.
2. Load a product.
3. Create or open a text panel.
4. With sanitization disabled, try to include an `iframe` or another element that sanitization would remove.
5. Ensure that the element is not removed.
6. Go back to the text panel and click the `sanitize` button so that it's enabled.
7. Try to include an `iframe`.
8. This time ensure that the `iframe` is removed correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/737)

### DIFF
--- a/src/components/panel-editors/text-editor.vue
+++ b/src/components/panel-editors/text-editor.vue
@@ -11,6 +11,7 @@
                 v-model="panel.content"
                 height="400px"
                 :left-toolbar="toolbarOptions"
+                :right-toolbar="rightToolbarOptions"
                 :toolbar="toolbar"
                 @fullscreen-change="onFullscreenChange"
             ></v-md-editor>
@@ -46,6 +47,7 @@ import { Options, Prop, Vue, Watch } from 'vue-property-decorator';
 import VueMarkdownEditor, { langMap } from '@/plugins/markdown-editor/index';
 import { TextPanel } from '@/definitions';
 import { applyTextAlign } from '@/utils/styleUtils';
+import { useUserStore } from '@/stores/userStore';
 import WETDashboardItemV from '../support/wet-dashboard-item.vue';
 import WETComponents from '../support/wet-component-templates.json';
 import DOMPurify from 'dompurify';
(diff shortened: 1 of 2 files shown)


RyanCoulsonCA (commented):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/737)* status: all files reviewed, 2 unresolved discussions (waiting on @spencerwahl)

___
*[a discussion](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/737#-OX94KJV-wBp_pQxX4d3:-OX9Z7008glYQpw48phY:bzbaqoi) (no related file):*

> This is pretty annoying, so maybe it's worth storing their preference in the user store for convenience.

Only downside I see to this is if its a single boolean controlling all textboxes someone might accidentally remove things they wanted to keep elsewhere in their product.
</blockquote></details>

Yeah that's a good point, and storing the state per-textbox would be a pain. It's probably best to leave it as is unless anyone has exceptions. 
___
*[`src/components/panel-editors/text-editor.vue` line 503 at r1](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/737#-OX93gQT4amQNnpAb_E-:-OX9ZVcKBkOrfdMBDXGE:b-fnwc0l) ([raw file](https://github.com/ramp4-pcar4/storylines-editor/blob/9d6e2264e7888ef080889d6a00934d32133c77f1/src/components/panel-editors/text-editor.vue#L503)):*

BLOCKING :cookie:
</blockquote></details>

:trophy:

IshavSohal (approved):
> *[Reviewable](https://reviewable.io/reviews/ramp4-pcar4/storylines-editor/737)* status: all files reviewed, 2 unresolved discussions (waiting on @spencerwahl)

